PR #153, opened 2026-01-17
136 ai generate flash card suggestions from highlight

Tumetsu:


frontend/src/pages/BookPage/FlashcardsTab/FlashcardCard.tsx
```diff
@@ -24,16 +24,24 @@ export interface FlashcardCardProps {
   onEdit: () => void;
   component?: React.ElementType;
   showSourceHighlight?: boolean;
+  variant?: 'default' | 'suggestion';
```
Tumetsu: Instead of using FlashcardCard.tsx for these Ai suggestion cards, lets just create a new component called FlashcardSuggestionCard.tsx. It should display the question and answer in a similar way as FlashcardCard.tsx but with dashed borders.

So remove these changes from the FlashcardCard.tsx file.

PR #207, opened 2026-02-17
Add highlight labels feature

Tumetsu:
## Summary

- **Promote HighlightStyle from JSON column to a dedicated `highlight_styles` table** with FK from highlights, supporting device color/style tracking and user-assignable labels with UI colors
- **Add highlight label management API** (`/highlight-labels`) with book-specific and global label CRUD, plus a 7-level priority resolution chain (combination+book > color-only+book > style-only+book > combination+global > color-only+global > style-only+global > none)
- **Regenerate frontend TypeScript types** with new `HighlightLabelInBook`, `HighlightLabelCreate`, `HighlightLabelUpdate` models and generated API client

## Test plan

- [x] 242 backend tests passing (24 new tests for labels feature + style resolver)
- [x] All linting and type checking clean (ruff, pyright)
- [x] Frontend types regenerated successfully from OpenAPI spec
- [ ] Manual testing of highlight upload with device_color/device_style
- [ ] Manual testing of label CRUD via highlight-labels endpoints
- [ ] Manual testing of label resolution priority chain

### DIFF
--- a/mcp-server/README.md
+++ b/mcp-server/README.md
@@ -99,6 +99,13 @@ crossbill-mcp
 - **tag_highlight** - Add a tag to a highlight
 - **untag_highlight** - Remove a tag from a highlight
 
+### Highlight Labels
+
+- **get_book_highlight_labels** - Get all highlight labels for a book with resolved labels and counts
+- **get_global_highlight_labels** - Get all global default highlight labels
+- **update_highlight_label** - Update label text and/or UI color on a highlight style
+- **create_global_highlight_label** - Create a new global default highlight label
+
 ### Flashcards
 
 - **get_flashcards** - Get all flashcards for a book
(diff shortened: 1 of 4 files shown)


backend/alembic/versions/041_add_highlight_style_to_highlights.py
```diff
@@ -15,29 +15,181 @@
...
+                "style": row.style,
+            },
+        )
+
+    # 5. Drop the old JSON column
     op.drop_column("highlights", "highlight_style")
+
+
+def downgrade() -> None:
+    op.add_column(
```
Tumetsu: This downgrade is also wrong, since production won't have the highlight_style in its database. So we do not want to revert to that. Instead just remove the foreign key introduced in this migration along with the table.

PR #214, opened 2026-03-03
UI redesign

Tumetsu:


frontend/src/pages/BookPage/navigation/MobileBottomNav.tsx
```diff
@@ -0,0 +1,42 @@
+import {
+  ChapterListIcon,
+  FlashcardsIcon,
+  HighlightsIcon,
+  ReadingSessionIcon,
+} from '@/theme/Icons.tsx';
+import { BottomNavigation, BottomNavigationAction, Paper } from '@mui/material';
+import { useNavigate, useParams, useRouterState } from '@tanstack/react-router';
+
+const getActiveTab = (pathname: string): string => {
```
Tumetsu: This is not a good name because we are not dealing with tabs anymore rather than with sub pages.

frontend/src/pages/BookPage/HighlightsTab/HighlightsTab.tsx
```diff
@@ -148,62 +183,176 @@ export const HighlightsTab = ({
...
+      handleLabelClick,
+      allHighlights,
+      handleBookmarkClick,
+    ]
+  );
+
   return (
     <>
-      {!isDesktop && (
+      {/* Desktop: portal left sidebar content */}
```
Tumetsu: Could we refactor this left side column to some subcomponent inside this file for better readability? Data could be passed in by props. SAme applies to otehr Tab views with similar structure.

frontend/src/routes/book.$bookId.tsx
```diff
@@ -1,21 +1,28 @@
 import { BookPage } from '@/pages/BookPage/BookPage';
 import { createFileRoute, redirect } from '@tanstack/react-router';
 
+type RedirectRoute =
+  | '/book/$bookId/highlights'
+  | '/book/$bookId/flashcards'
+  | '/book/$bookId/sessions'
+  | '/book/$bookId/structure';
+
+const tabRouteMap: Partial<Record<string, RedirectRoute>> = {
```
Tumetsu: This route mapping is now duplicated here, in DesktopNavLinks.tsx and in another format in the MobileBottomNav.tsx. That is not viable. These should be defined in a single place such as routes.tsx etc. where all of these places import the routes and their types.

PR #410, opened 2026-07-18
Add unlinking of note links to highlights and chapters

Tumetsu:
## Summary

Notes could be linked to highlights and chapters, but there was no UI to remove a link. This adds unlinking in three places, plus note linking in the chapter dialog:

- **Note view modal**: each linked highlight and chapter in the tabs gets an unlink icon (chapter rows via `ListItem` secondary action, highlight cards via an overlay button).
- **Highlight modal**: each linked note card gets an unlink icon.
- **Chapter dialog**: each linked note card gets an unlink icon, and the notes tab gains the same "Link existing note" button the highlight modal has.

Unlinking is one click with a snackbar — no confirmation dialog, since links are easy to restore.

## Implementation notes

- No backend changes: links are replaced wholesale via `PUT /notes/{id}`, so a new shared `useNoteLinks` hook re-sends the note with the target id added/removed and centralizes query invalidation and snackbars. `HighlightNotesSection`'s local link mutation is replaced by the hook.
- `NoteCard` gains a right-edge `action` slot (and its keydown handler now ignores events bubbling from the action button).
- `NotePickerDialog` moved from the highlight modal folder to shared `Notes/components/` with its title as a prop, so the chapter dialog can reuse it.

## Testing

- `tsc --noEmit` and `eslint --max-warnings 0` pass (no frontend test suite).
- Worth a manual pass: unlink from all three surfaces, confirm lists/tab counts refresh and clicking the icon doesn't open the underlying modal.

### DIFF
--- a/frontend/src/pages/BookPage/Notes/components/NoteTabs.tsx
+++ b/frontend/src/pages/BookPage/Notes/components/NoteTabs.tsx
@@ -1,7 +1,19 @@
 import type { Highlight, NoteLinkedChapter, NoteWithLinks } from '@/api/generated/model';
 import { HighlightCard } from '@/pages/BookPage/Highlights/HighlightCard.tsx';
 import { NoteFlashcardSection } from '@/pages/BookPage/Notes/components/NoteFlashcardSection.tsx';
-import { Box, List, ListItemButton, ListItemText, Stack, Tab, Tabs } from '@mui/material';
+import { LinkOffIcon } from '@/theme/Icons.tsx';
+import {
+  Box,
+  IconButton,
+  List,
+  ListItem,
+  ListItemButton,
+  ListItemText,
+  Stack,
+  Tab,
+  Tabs,
+  Tooltip,
+} from '@mui/material';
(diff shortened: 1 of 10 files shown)


frontend/src/pages/BookPage/Notes/components/NoteTabs.tsx
```diff
@@ -40,9 +56,22 @@ export const NoteTabs = ({
       content: (
         <Stack component="ul" sx={{ gap: 2, listStyle: 'none', p: 0, m: 0 }}>
           {highlights.map((highlight) => (
-            <li key={highlight.id}>
+            <Box component="li" key={highlight.id} sx={{ position: 'relative' }}>
               <HighlightCard highlight={highlight} onOpenModal={onOpenHighlight} />
-            </li>
+              {onUnlinkHighlight && (
+                <Tooltip title="Unlink from note">
```
Tumetsu: I think this component could be separated and reused in different places right? To avoid copy paste.

PR #411, opened 2026-07-18
Unify the book-page detail modals (chapter, highlight, note)

Tumetsu:
## Summary

The three detail modals on BookPage — ChapterDetailDialog, HighlightViewModal and NoteViewModal — share the same shape (CommonDialog → content → toolbar → tabs of linked items) but drifted apart over time. This PR extracts the shared pieces, aligns the visual logic, and brings the note modal up to feature parity. Net **−140 lines** while adding two features.

Each concern is a separate commit for easier review:

1. **Extract shared dialog and list primitives** — `DialogToolbar`, `CardList`, `EmptyStateText`, `copyUrlWithSearchParam`; moves `ProgressBar` to `components/dialogs` (fixes a cross-feature import); aligns footer `justifyContent`.
2. **Unify modal tabs with a shared `DialogTabs` component** — replaces three divergent implementations and their duplicated `formatTabLabel` helpers. Touch-event stopPropagation is now built in (previously missing in NoteTabs, so tab gestures fought swipe navigation). *Behavior change:* the note modal now always shows all tabs with counts and empty states instead of hiding empty ones, matching the other modals.
3. **Deduplicate linked-notes sections into `LinkedNotesSection`** — merges the near-identical `HighlightNotesSection` and chapter `NotesSection`, parameterized by link target. The chapter dialog's Notes tab now shows a count (query lifted into the dialog, so it loads on dialog open instead of first tab click).
4. **Add a copy-link action to the note toolbar** — mirrors the highlight toolbar; URL is built on the notes route explicitly since `noteId` is only validated there.
5. **Add prev/next navigation to the note view modal** — arrows, arrow keys, swipe, progress bar and fade transition, mirroring the highlight modal. Follows the filtered list on NotesPage, replace-navigates so Back still closes the modal, suspends navigation while editing, and resets edit state when the note changes.

## Testing

- `npm run type-check` and `npm run lint` pass on **every commit** (verified per commit).
- Manually exercised in the app: tab counts/switching in all three modals, link/unlink/create note from the highlight modal and chapter dialog, note copy-link deep link, and note prev/next navigation.


Tumetsu:
Added four follow-up commits, one per topic:

- **Fix highlight copy-link from non-highlights routes** — links copied from a highlight opened inside the chapter dialog pointed at the structure route, where `highlightId` is stripped by `validateSearch`, so the pasted link opened nothing. Now built on the highlights route explicitly, same pattern as the note copy-link.
- **Use theme tokens in ProgressBar** — the fill hardcoded `#43311E` (with a comment admitting it's `primary.main`) and the track hardcoded the exact rgba value that already existed in the theme as `customColors.backgrounds.subtle`.
- **Use a theme token for assistant chat bubbles** — replaces the hardcoded `grey.100` and removes a dead `grey.800` dark-mode branch (the app theme is light-only).
- **Migrate remaining card lists to CardList** — the three mechanical matches (FlashcardSuggestions, FlashcardChapterList, HighlightsList) plus `aria-label` support on CardList. ReadingSessions/ChapterNav/BookmarkList/TagChipList intentionally left alone — different spacing or list semantics.

Type-check and lint verified on each commit individually.

PR #447, opened 2026-07-24
Harden auth and rate limiting

Tumetsu:


Tumetsu:
Some of the commits seem a bit fudged with cross contaminated by changes from other commits and the recent Matt Pocock skill experiment. I'll leave them be since implementation seems good enough. Have to keep eye on that this won't turn into pattern in the future with agents.

PR #484, opened 2026-07-31
Implement refresh token rotation with grace window

Tumetsu:
## Summary

Implements secure refresh token rotation with a configurable grace window to handle network failures and race conditions during token refresh. When a refresh token is rotated, the old token is marked as spent but remains valid for a brief window, allowing clients that didn't receive the rotation response to retry and get the replacement token instead of being logged out.

### Backend

**Token Rotation Logic**
- Modified `RefreshAccessTokenUseCase` to distinguish between rotated tokens (with a successor) and revoked tokens (logout/password change)
- Added `replaced_by_jti` field to `RefreshToken` entity to track the successor token
- Implemented `successor_within(grace_window)` method to check if a spent token's replacement is still available
- Added `rotate()` method to repository to atomically spend a token and store its successor

**Grace Window Handling**
- Added `REFRESH_TOKEN_ROTATION_GRACE_SECONDS` configuration (default 60 seconds)
- When a revoked token is presented:
  - If its replacement is still within the grace window, return the replacement (handles lost responses and races)
  - If outside the grace window, treat as replay and revoke the entire family (security)
  - If revoked without a successor (logout), always reject (final revocation)

**Database**
- Added migration `060_add_replaced_by_jti_to_refresh_tokens.py` to add the `replaced_by_jti` column
- Updated ORM model and mapper to handle timezone-aware datetime values

**Token Service**
- Extracted `reissue_token_pair()` to rebuild token pairs for already-issued refresh tokens (used when returning a cached successor)

### Frontend

**Session Synchronization**
- Refactored `AuthContext` to use a single `syncSession()` function that handles all refresh scenarios (mount, scheduled tick, focus return)
- Implemented `refreshSession()` with single-flight deduplication: concurrent refresh triggers share one request to avoid presenting the same token twice
- Added transient retry logic (500ms, 2s delays) for network failures, distinct from the longer backoff for persistent failures

**Refresh Scheduling**
- Refresh scheduled at 75% of token lifetime (leaving room for retries before expiry)
- On app focus, refresh immediately if token is within 60 seconds of expiry (handles iOS PWA suspension)
- Separate retry backoff (5s, 15s, 60s) for persistent failures, allowing the session to survive transient network issues

**Error Handling**
- New `SessionExpiredError` class to distinguish server rejection of the refresh token (session over) from transient failures (network, 5xx)
- Only `SessionExpiredError` triggers logout; other failures are retried
- Response interceptor only retries non-auth endpoints (auth endpoints answer 401 about credentials themselves)

### Testing

- Added comprehensive end-to-end test suite `test_refresh_token_rotation.py` covering:
  - Normal rotation and cookie updates
  - Lost response recovery (retry gets the same successor)
  - Replay detection after grace window expires
  - Multi-generation replay detection (always rejected)
  - Logout finality (no successor even within grace window)
  - Successor link recording in database
  - Expired token cleanup on refresh

## Notable Implementation Details

- **Atomic rotation**: `rotate()` method updates the old token's `replaced_by_jti` and saves the new token in a single transaction
- **Family revocation**: Replays outside the grace window revoke the entire token family, not just the presented token
- **Single-flight refresh**: Frontend deduplicates concurrent refresh attempts to prevent presenting the same token twice (which would trigger replay detection)
- **Graceful degradation**: Network failures don't log users out; only server rejection of the refresh token itself ends the session
- **Timezone handling**: Mapper restores UTC timezone info that SQLite drops, ensuring domain layer comparisons work correctly

https://claude.ai/code/session_015LUJaF46u5LtJ9XZPedH1S


frontend/src/api/axios-instance.ts
```diff
@@ -11,24 +11,98 @@ export const AXIOS_INSTANCE = Axios.create({
...
-}> = [];
-
-const processQueue = (error: unknown, token: string | null = null) => {
-  failedQueue.forEach((promise) => {
-    if (error) {
-      promise.reject(error);
-    } else {
-      promise.resolve(token);
-    }
+/**
```
Tumetsu, line 14: Comments like this which explain some past incident are not necessary. Instead just leave explanation what the code does/why.

Issue #533, opened 2026-08-03
Semantic search: remaining work after #532


Tumetsu:
#532 lands the semantic-search infrastructure, backfill ingestion and the read side. It deliberately stops short of several things so that each can land as its own reviewable change. This is the ledger of what is left, in the order it should be done.

Until item 1 lands, **new and edited content is not searchable until someone runs a backfill**. That is the main functional gap.

## 1. Live enqueue-on-write (phase D)

The `EmbeddingEnqueuer` application service, called from note create/update, highlight upload and chapter-digest generation, so content is embedded as it is written.

- No-ops when the feature flag is off, and **swallows enqueue failures** — a missed embedding must never fail a user's write; the backfill sweep reconciles it.
- Single edits enqueue a bare job; a highlight upload enqueues a `JobBatch`.
- This is the first change that makes a source module depend on a semantic port, so it is where that coupling should be argued.
- The worker task's `batch_id` is currently required. A bare per-edit job has no batch, so this widens it back to `int | None` — a one-line change, but worth doing deliberately rather than by accident.

Reference implementation exists on the `semantic-search-embeddings` branch (commit `7207ada7`).

## 2. Explicit soft-delete cleanup

Remove a soft-deleted highlight's embedding at the delete site instead of waiting for the next backfill sweep. Small, and disproportionately worth care:

- The ids in a delete request are **unverified caller input**. The cleanup must be driven by the ids the repository *actually* deleted, not the ids that were asked for. Doing it the obvious way lets one user prune another user's embeddings — this was found in review on the old branch, reproduced live twice.
- That means `soft_delete_by_ids` changes from returning a count to returning the ids it deleted. A `user_id` filter on the delete alone is **not** sufficient: it does not cover ids of your own from a different book.
- The cleanup must not fail the request. The soft delete has already committed and the two are not one transaction, so an exception here would 500 a delete that succeeded.

Reference: commits `7254ce00` and `21b341e7` on `semantic-search-embeddings`.

## 3. Frontend

Surfacing search and related content in the UI. The generated API client is already committed (CI enforces it is current); it simply has no callers yet.

## 4. Production wiring

- OpenRouter `baai/bge-m3`
- Confirm pgvector is available on Railway and run migration `062` there
- **Check whether `EMBEDDING_DIMENSIONS` is set in the Railway environment.** It is no longer a setting, and `Settings` forbids extra inputs, so if it is set the app will fail to boot after this merges. This bit locally already.

[...]

PR #547, opened 2026-08-05
Embed content automatically on create and edit

Tumetsu:
Item 1 of #533 — phase D part 1. Until now, new and edited content was invisible to semantic search until someone pressed backfill. This closes that gap for all three content types: notes, highlights and digests.

## The seam

An `EmbeddingEnqueuer` application service behind `EmbeddingEnqueuerProtocol`, called from four write use cases:

| Path | Call |
|---|---|
| note create / update | `enqueue_for(NOTE, …)` — bare job, no batch |
| highlight upload | `enqueue_many(HIGHLIGHT, …)` — tracked batch of slices |
| chapter-digest generation | `enqueue_for(DIGEST, …)` |

It is deliberately weak in both directions: it no-ops when no embedding provider is configured, and it swallows enqueue failures. A queue that is down must not cost a user their note; the backfill sweep reconciles whatever is missed.

This is the first source-module → semantic dependency, so the narrowness is the argument for it: `notes` and `reading` commands import the port and the `ContentType` enum, never the service, the repository or the client. No import-linter contract changes; all five still pass.

Every call site enqueues **after** the repository save. Repositories commit inside `save()`, so a job that starts running immediately still reads committed content rather than the row it is replacing.

A single edit enqueues a bare job — a `JobBatch` per note save would be a batch row per edit, each of one job — which is why the worker task's `batch_id` widens back to `int | None`. `after_process` already tolerated its absence. A highlight upload enqueues only the ids `bulk_save` actually produced: a KOReader sync resends the whole book, so enqueuing the deduplicated ids would re-embed a book on every sync.

Answering a digest's questions does not re-enqueue; only summary and keypoints are embedded.

## Three deviations worth reviewing

**The reference implementation on `semantic-search-embeddings` (`7207ada7`) did not survive contact.** It predates #532's slice model and enqueues `generate_content_embedding` with a single `content_id`. The merged task takes `content_ids: list[int]`. Everything enqueue-shaped was rewritten against the merged signature.

**Failure accounting deviates from that reference on purpose.** It shrank `total_jobs` to what was enqueued; this uses the merged backfill's `mark_job_failed()` instead, so a batch that broke partway cannot terminate as "completed" with nothing to say the rest were dropped. That is what made a shared `record_dropped_slices` possible — so the "decide before extracting" question in #533 no longer applies to the two semantic paths, only to the digest path, which still diverges.

**jscpd did see the third-copy clone** that #533 predicted it could not. Extracting the accounting tail removed it. Ratchet lowered 2.66 → 2.6 (measured 2.5972), in the same change.

The first commit is that extraction on its own, with no behaviour change; the second is the feature.

## Tests

13 API tests in `test_semantic_live_enqueue.py`, through the endpoints. Two fixture changes are load-bearing:

- **The `client` fixture now installs the contract-checked queue fake.** Without it, every write path 500s at DI resolution in tests — before the enqueuer's swallowing can help. The fake also binds each enqueue's kwargs to the real worker task, which is exactly the check that catches `content_id=` against a task taking `content_ids=`. `test_semantic_backfill`'s local fixture folded into a shared `job_queue` one, so there is exactly one fake per test.
- **The write-path gate is `Settings.embeddings_enabled` on the container, not the `is_embeddings_enabled()` the router tests patch.** Different lever — patching the wrong one passes while asserting nothing. `embeddings_enabled()` / `embeddings_disabled()` force it explicitly in both directions, because the ambient value differs between a developer with `EMBEDDING_PROVIDER` in a `.env` and CI. Full suite run both ways: 758 pass in each.

All five call sites were falsifiability-checked — gate removed, slicing removed, swallowing removed, and each enqueue call deleted in turn — each failing exactly the tests that should fail.

This also adds the first API coverage of `POST /chapters/{id}/digest/generate`, which had none.

## Still open on #533

Item 2 (soft-delete cleanup, with the cross-tenant bug), frontend, production wiring, and the digest path's failure-accounting divergence.

### DIFF
--- a/backend/src/application/notes/commands/create_note_use_case.py
+++ b/backend/src/application/notes/commands/create_note_use_case.py
@@ -77,8 +77,6 @@ class CreateNoteUseCase:
         )
         note = await self.note_repository.save(note)
 
-        # After the save, never before: the repository commits, so a job picked
-        # up the instant it is enqueued still reads the note that was written.
         await self._embedding_enqueuer.enqueue_for(ContentType.NOTE, note.id.value, user_id)
 
         logger.info("created_note", note_id=note.id.value, book_id=book_id)
--- a/backend/src/application/notes/commands/create_note_use_case.py
+++ b/backend/src/application/notes/commands/create_note_use_case.py
@@ -8,6 +8,8 @@ from src.application.library.protocols.chapter_repository import ChapterReposito
 from src.application.notes.commands.helpers import parse_note_kind, validate_link_targets
 from src.application.notes.protocols.note_repository import NoteRepositoryProtocol
 from src.application.reading.protocols.highlight_repository import HighlightRepositoryProtocol
+from src.application.semantic.content_type import ContentType
+from src.application.semantic.protocols.embedding_enqueuer import EmbeddingEnqueuerProtocol
 from src.application.tagging.protocols.tag_repository import (
     TagRepositoryProtocol,
 )
@@ -27,12 +29,14 @@ class CreateNoteUseCase:
         chapter_repository: ChapterRepositoryProtocol,
         highlight_repository: HighlightRepositoryProtocol,
         tag_repository: TagRepositoryProtocol,
+        embedding_enqueuer: EmbeddingEnqueuerProtocol,
     ) -> None:
         self.note_repository = note_repository
         self.book_repository = book_repository
         self.chapter_repository = chapter_repository
         self.highlight_repository = highlight_repository
         self.tag_repository = tag_repository
+        self._embedding_enqueuer = embedding_enqueuer
 
     async def create_note(
         self,
@@ -72,6 +76,10 @@ class CreateNoteUseCase:
             tag_ids=tag_ids,
         )
         note = await self.note_repository.save(note)
+
+        # After the save, never before: the repository commits, so a job picked
+        # up the instant it is enqueued still reads the note that was written.
+        await self._embedding_enqueuer.enqueue_for(ContentType.NOTE, note.id.value, user_id)
 
         logger.info("created_note", note_id=note.id.value, book_id=book_id)
         return note
--- a/backend/src/application/notes/commands/update_note_use_case.py
+++ b/backend/src/application/notes/commands/update_note_use_case.py
@@ -6,6 +6,8 @@ from src.application.library.protocols.chapter_repository import ChapterReposito
 from src.application.notes.commands.helpers import parse_note_kind, validate_link_targets
 from src.application.notes.protocols.note_repository import NoteRepositoryProtocol
 from src.application.reading.protocols.highlight_repository import HighlightRepositoryProtocol
+from src.application.semantic.content_type import ContentType
+from src.application.semantic.protocols.embedding_enqueuer import EmbeddingEnqueuerProtocol
 from src.application.tagging.protocols.tag_repository import (
     TagRepositoryProtocol,
 )
@@ -25,11 +27,13 @@ class UpdateNoteUseCase:
         chapter_repository: ChapterRepositoryProtocol,
         highlight_repository: HighlightRepositoryProtocol,
         tag_repository: TagRepositoryProtocol,
+        embedding_enqueuer: EmbeddingEnqueuerProtocol,
     ) -> None:
         self.note_repository = note_repository
         self.chapter_repository = chapter_repository
         self.highlight_repository = highlight_repository
         self.tag_repository = tag_repository
+        self._embedding_enqueuer = embedding_enqueuer
 
     async def update_note(
         self,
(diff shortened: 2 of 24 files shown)


backend/src/application/notes/commands/create_note_use_case.py
```diff
@@ -73,5 +77,9 @@ async def create_note(
         )
         note = await self.note_repository.save(note)
 
+        # After the save, never before: the repository commits, so a job picked
```
Tumetsu: remove this comment

backend/src/application/notes/commands/update_note_use_case.py
```diff
@@ -67,5 +71,9 @@ async def update_note(
         )
         note = await self.note_repository.save(note)
 
+        # After the save, so the re-embedding reads the edited text rather than
```
Tumetsu: remove this comment

backend/src/application/reading/commands/chapter_digest/generate_chapter_digest_use_case.py
```diff
@@ -119,4 +123,13 @@ async def generate_digest(self, chapter_id: ChapterId, user_id: UserId) -> Chapt
             chapter_id=chapter_id.value,
             keypoints_count=len(ai_result.keypoints),
         )
-        return await self.digest_repo.save(entity)
+        saved = await self.digest_repo.save(entity)
+
+        # The digest's id only exists after the save, and its text is what gets
```
Tumetsu: remove this comment

backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
```diff
@@ -201,7 +206,16 @@ async def upload_highlights(
 
         # Step 5: Bulk save unique highlights
         if unique:
-            await self.highlight_repository.bulk_save(unique)
+            saved = await self.highlight_repository.bulk_save(unique)
+            # Only the ids the save actually produced -- the duplicates were
```
Tumetsu: remove this comment

backend/src/application/semantic/protocols/embedding_enqueuer.py
```diff
@@ -0,0 +1,29 @@
...
+calling write path -- a failure to enqueue must not fail a note save or a
+highlight upload.
+"""
+
+from typing import Protocol
+
+from src.application.semantic.content_type import ContentType
+
+
+class EmbeddingEnqueuerProtocol(Protocol):
```
Tumetsu, line 13: Hmm, do we now have different ways/structure on how the embedding is triggered for backfill batch processes and for these live embeddings? Looks like backend/src/application/semantic/commands/enqueue_content_embeddings_use_case.py uses backend/src/application/semantic/batching.py for triggering the embeddings. 

Why is batching repo save in the application layer and not behind port? Is there an inconsistency here?

backend/src/application/semantic/services/embedding_enqueuer.py
```diff
@@ -0,0 +1,149 @@
...
+from src.domain.jobs.entities.job_batch import JobBatch, JobBatchType
+
+logger = structlog.get_logger(__name__)
+
+_TASK = "generate_content_embeddings"
+_RETRIES = 3
+_TIMEOUT_SECONDS = 300
+
+
+class EmbeddingEnqueuer:
```
Tumetsu, line 28: Hmm, why is this in the application layer and not behind port which is implemented in the infra layer. Does this break DDD pattern we have?

Tumetsu: Looks like this file implements the port with same name, but now both of these are located in the application layer.

backend/src/application/semantic/services/embedding_enqueuer.py
```diff
@@ -51,14 +54,7 @@ async def enqueue_for(self, content_type: ContentType, content_id: int, user_id:
...
         try:
-            await self._queue_service.enqueue(
-                _TASK,
-                retries=_RETRIES,
-                timeout_seconds=_TIMEOUT_SECONDS,
-                content_type=content_type.value,
-                content_ids=[content_id],
-                user_id=user_id,
-            )
+            await enqueue_embedding_job(content_type, [content_id], user_id, self._queue_service)
```
Tumetsu, line 51: Explain what is the difference between enqueue job and batch

backend/src/worker.py
```diff
@@ -255,7 +255,18 @@ class EmbeddedWorker(Worker[Context]):
 
 
 def create_embedded_worker(queue: Queue, concurrency: int = 2) -> EmbeddedWorker:
-    """Create a Worker instance for running inside the app process."""
+    """Create a Worker instance for running inside the app process.
+
+    Binds the app's already-connected queue as this module's queue. A task that
```
Tumetsu: Why is this whole thing needed? Is it because worker might have different DI container than the calling app?

Issue #599, opened 2026-08-21
Server: is_new upload flag — deliberate re-highlights revive removed or deleted highlights


Tumetsu:
Blocked by: #596

Part 4 of the **highlight deletion sync** feature. Today the upload dedup check includes soft-deleted rows so that push-only clients can't resurrect web deletions (`highlight_repository.get_existing_hashes`) — with the side effect that a passage once deleted on the web can never be highlighted again. Under bidirectional sync the plugin can tell the server which pushed highlights are genuinely **new on this device** (absent from its last-pulled snapshot), so the server can distinguish a deliberate re-highlight from a stale echo without any clock comparison.

## Scope

- `HighlightCreate` (`backend/src/infrastructure/reading/schemas/highlight_schemas.py`) gains optional `is_new: bool = false`.
- `HighlightUploadUseCase` duplicate resolution (`backend/src/application/reading/commands/highlights/highlight_upload_use_case.py`) becomes:
  - matches a **live visible** row → existing edit-merge, unchanged;
  - matches a **removed-from-devices** row (#596) and `is_new` → clear `removed_from_devices_at`, apply incoming note/style edits;
  - matches a **soft-deleted** row and `is_new` → `restore()` (exists unused on the entity) + apply edits — its flashcards stay gone, they were hard-deleted at web-delete time;
  - **unflagged** duplicate of a removed or deleted row → skipped, exactly as today (protects against stale devices and legacy plugin versions).
- Note the required repo change: the reconciliation path currently loads only live rows (`find_live_by_content_hashes`); flagged-new incoming highlights need matching against removed/deleted rows too.
- Ordering within the use case: removals (`removed_ids`, #598) are evaluated before duplicate resolution, so an `is_new` push in the same request wins over a removal of the same text — deterministic either way.

## Acceptance criteria

- `is_new` push revives a removed highlight (reappears in the ereader pull) and restores a web-deleted one.
- Unflagged pushes of the same texts change nothing — legacy behavior byte-for-byte.
- Within-batch dedup and the existing last-writer-wins note merge unaffected.
- API tests per `writing-tests`; pyright + suite green.

Note: #598 modifies the same use case — land it first and rebase this on top.

PR #605, opened 2026-08-21
Server: accept removed_ids in the highlight upload request (#598)

### LINKED ISSUE #598: Server: accept removed_ids in the highlight upload request (remove from devices)
Blocked by: #596

Part 3 of the **highlight deletion sync** feature. The channel through which the KOReader plugin propagates a device-side deletion: it marks highlights *removed from devices* (#596) — it does **not** soft-delete them and does **not** touch flashcards or bookmarks (unlike the existing web delete cascade, which is deliberately not reused here).

**Design revision (D7)**: removals ride *inside* the existing upload request rather than a separate endpoint — one round trip on flaky e-reader WiFi, and one server transaction so a sync can't die between "removals applied" and "highlights uploaded".

## Scope

- `HighlightUploadRequest` (`backend/src/infrastructure/reading/schemas/highlight_schemas.py`) gains optional `removed_ids: list[int] = []`.
- `HighlightUploadUseCase` (`backend/src/application/reading/commands/highlights/highlight_upload_use_case.py`) evaluates removals **first** — setting `removed_from_devices_at` on ids that are owned by the user, belong to the resolved book, and are live — then runs the regular upload/dedup logic, all in the same transaction. (Ordering matters for the pathological case where a removed id's text hash also appears in the pushed set: removal first, then dedup, keeps the outcome deterministic — unflagged duplicate stays removed; a later `is_new` duplicate, once #599 lands, revives.)
- **Idempotent**: ids already removed, already soft-deleted, or unknown/foreign are ignored, not errors (the plugin re-sends removals after an interrupted sync).
- Response (`HighlightUploadResponse`) gains `highlights_removed: int`.
- Legacy plugins omit the field → byte-for-byte today's behavior.

## Acceptance criteria

- An upload with `removed_ids` marks the rows and the ereader pull (#596) no longer returns them; web reads still do.
- Flashcards, bookmarks, and embeddings of removed highlights are untouched.
- Repeating the same request succeeds with no further effect; uploads without the field are unchanged.
- API tests per the `writing-tests` policy; pyright + suite green.

Note: #599 modifies the same use case — land this first, #599 rebases on it.

Tumetsu:
Fixes #598

Part 3 of the highlight deletion sync. A highlight the reader deletes on an e-reader now propagates to the server — inside the upload request the device already sends, not through an endpoint of its own.

## What changed

| Layer | Addition |
|---|---|
| `HighlightUploadRequest` | optional `removed_ids: list[int] = []` |
| `HighlightUploadUseCase` | `_remove_deleted_from_devices()`, called as step 2 |
| `HighlightRepository` | `mark_removed_from_devices()` — set-based, cascades to nothing |
| `HighlightUploadResponse` | `highlights_removed: int` |

`upload_highlights` gains exactly one line of deletion logic:

```python
removed_count = await self._remove_deleted_from_devices(removed_ids or [], user_id_vo, book_id)
```

Everything else about deletion — which IDs are valid, why unknown ones are ignored, why nothing cascades — lives in the two named functions, so the upload flow itself reads as it did. The return type is now `HighlightUploadResult(created, skipped, removed_from_devices)` rather than a tuple that would have grown to three unlabelled ints.

**Idempotent by predicate**: the marking query is scoped by user, book, `deleted_at IS NULL` and `removed_from_devices_at IS NULL`, and reports back only the rows it actually marked. Unknown, foreign, soft-deleted and already-withheld IDs are skipped, not rejected — the plugin re-sends its removals after an interrupted sync.

**Ordering**: removals are applied before anything is built or deduplicated. When a withheld highlight's own text arrives in the same push — as it does whenever the device re-sends its whole set — the removal is already in place and deduplication skips the push as the duplicate it is. `get_existing_hashes` counts withheld rows, so nothing is recreated.

## Acceptance criteria

| Criterion | Covered by |
|---|---|
| `removed_ids` marks the rows; the e-reader pull drops them, web reads keep them | `test_removed_highlight_leaves_the_pull_but_stays_on_the_web` |
| Flashcards and bookmarks untouched | `test_flashcards_and_bookmarks_of_a_removed_highlight_survive` |
| Repeating the request has no further effect | `test_resending_the_same_removal_changes_nothing` |
| Unknown / soft-deleted IDs ignored | `test_unknown_and_soft_deleted_ids_are_ignored` |
| Another user's highlight untouched | `test_another_users_highlight_is_left_alone` |
| Uploads without the field unchanged | `test_upload_without_removed_ids_removes_nothing` |

Embeddings are untouched because the removal path never reaches the enqueuer; asserting that would mean asserting on a mock, which the test policy rules out.

## Verification

- `uv run pytest` — 842 passed (was 836).
- `make lint` — ruff clean, pyright 0 errors, import-linter 5 contracts kept.
- jscpd — 113 clones, 2.4% before and after; threshold left at 2.55.
- `make api-client` — the regenerated frontend client picks up both fields; no consumers to break.

Each new test was proven able to fail. Four mutations, each caught by exactly the intended test, then reverted: removal made a no-op (3 tests fail), user/book scope dropped from the query (isolation test), `deleted_at` filter dropped, `removed_from_devices_at` filter dropped.

## Deviation from the ticket

The ticket asks for removals and upload in **one server transaction**. `mark_removed_from_devices` commits on its own, like every other bulk write on this repository — there is no unit-of-work seam, and omitting the commit would lose removals on a removals-only request, since no later step commits when the push is empty.

So if the upload raises after removals land, the removals persist. The plugin re-sends them, they are idempotent, and the end state converges. Making it genuinely atomic means introducing transaction management across the repository, which is a redesign rather than this ticket.

Note: #599 modifies the same use case and rebases on this.

https://claude.ai/code/session_01EpipxtqKFVvxnGLm86Ts3y

### DIFF
--- a/backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
+++ b/backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
@@ -123,12 +123,6 @@ class HighlightUploadUseCase:
         7. Fills in the xpoints and position of duplicates stored without them
         8. Bulk saves unique highlights
 
-        The deletions ride inside the upload rather than in a call of their own,
-        so a sync is one round trip on flaky e-reader WiFi. They are applied
-        before deduplication runs: should a removed highlight's text also arrive
-        in the pushed set, the removal is already in place and deduplication
-        skips the push as the duplicate it is, leaving the highlight withheld.
-
         Args:
             client_book_id: Book identifier from client
             highlight_data_list: List of highlight data to upload
@@ -247,7 +241,7 @@ class HighlightUploadUseCase:
         )
 
         # Steps 6-7: reconcile the live rows the duplicates matched, loaded once
-        stored_duplicates = await self.highlight_repository.find_live_by_content_hashes(
+        stored_duplicates = await self.highlight_repository.find_reconcilable_by_content_hashes(
             user_id_vo, book_id, [duplicate.content_hash for duplicate in duplicates]
         )
         await self._sync_device_edits_of_duplicates(duplicates, stored_duplicates, book_id)
@@ -292,11 +286,6 @@ class HighlightUploadUseCase:
         only marks the highlight withheld and leaves the web copy whole. This is
         deliberately not the web's delete cascade.
 
-        Ids that are unknown, another user's, from another book, already
-        soft-deleted or already withheld are skipped rather than rejected: the
-        plugin re-sends its removals after an interrupted sync, and the second
-        attempt must succeed as quietly as the first.
-
         Args:
             removed_ids: Highlight IDs the device reports as deleted, unverified
             user_id: User the upload belongs to
@@ -308,8 +297,12 @@ class HighlightUploadUseCase:
         if not removed_ids:
             return 0
 
+        # Repeats cost a bind parameter each and mark nothing extra; the list is
+        # unbounded caller input, so collapse them before it reaches the query.
+        unique_ids = list(dict.fromkeys(removed_ids))
+
         removed = await self.highlight_repository.mark_removed_from_devices(
-            [HighlightId(highlight_id) for highlight_id in removed_ids], user_id, book_id
+            [HighlightId(highlight_id) for highlight_id in unique_ids], user_id, book_id
         )
 
         logger.info(
--- a/backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
+++ b/backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
@@ -57,6 +57,15 @@ class HighlightUploadData:
     koreader_note: str | None = None
 
 
+@dataclass(frozen=True)
+class HighlightUploadResult:
+    """What one sync did: the highlights it stored, skipped and withheld."""
+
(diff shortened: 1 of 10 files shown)


backend/src/application/reading/commands/highlights/highlight_upload_use_case.py
```diff
@@ -99,27 +108,36 @@ async def upload_highlights(
...
-        7. Bulk saves unique highlights
+        2. Withholds the highlights the reader deleted on a device
+        3. Batch fetches chapters by chapter_number
+        4. Creates domain entities using factory methods
+        5. Deduplicates using domain service
+        6. Applies the e-reader's newer note and style edits to skipped duplicates
+        7. Fills in the xpoints and position of duplicates stored without them
+        8. Bulk saves unique highlights
+
+        The deletions ride inside the upload rather than in a call of their own,
```
Tumetsu: This explanation is kinda meaningless. This is a sync method (will be renamed later) so deletion of deleted or removed highlights is a natural part of the process. The numbered list above already makes intention clear.

PR #623, opened 2026-08-23
Cap chapter text sent to digest, quiz and chat

### LINKED ISSUE #613: Cap the chapter text quiz and chat send to the model
Part of #593.

## Problem

Quiz and chat send the **entire extracted chapter text** to the model with no cap:

- `start_quiz` — `backend/src/infrastructure/ai/ai_service.py:111` builds the prompt as `f"...--- CHAPTER CONTENT ---\n{chapter_content}"`
- `seed_chat_context` — `backend/src/infrastructure/ai/ai_service.py:130` does the same and stores it as the opening user turn

`generate_digest` (`ai_service.py:72`) already caps its input at `content[:10000]`; quiz and chat have no equivalent.

For an EPUB whose nav data has no chapter delimiters, one "chapter" spans the whole book (#593: Keegan's *The First World War*, 1046 pages, one chapter named "Start"). Starting a quiz on it ships several hundred thousand characters in a single request.

Chat is worse than a one-off: `seed_chat_context` persists that prompt into `AIChatSession.message_history`, and `continue_chat` replays the whole history on every turn. One oversized seed is re-sent with every message the reader writes.

## Scope

Cap the chapter text that quiz and chat send, so a missing-delimiter book cannot balloon into an unbounded request. Purely a guardrail — no UI, no detection heuristics, no change to how chapters are derived.

Suggested shape:

- One shared cap constant + truncation helper in the AI infrastructure layer, applied in `start_quiz` and `seed_chat_context`, with `generate_digest` moved onto the same helper. Precedent for a single truncation point: `backend/src/infrastructure/semantic/content/content_source.py:50`.
- Head truncation (same strategy as digest) unless there is a reason to do better.
- Log when truncation happens (chapter id + original length), so oversized chapters are visible in the logs.

## Acceptance criteria

- Starting a quiz on a chapter whose extracted text exceeds the cap sends at most the cap to the model.
- Starting a chat on the same chapter persists a `message_history` whose seed prompt is at most the cap.
- A chapter shorter than the cap is sent unchanged.
- Truncation emits a log line naming the chapter and the dropped length.

jdbway:
## Summary

- Digest already truncated its input to 10,000 characters; quiz and chat had no cap at all, so a chapter with no real chapter boundaries (the whole book parsed as one chapter, e.g. when an EPUB's nav data is missing) sent the entire book as context on every turn.
- Adds a shared `truncate_chapter_content` helper in `AIService` that head-truncates at the same 10,000-char figure and logs the chapter ID + dropped character count when it fires.
- Routes `generate_digest`, `start_quiz`, and `seed_chat_context` through it. `seed_chat_context` needed a new `chapter_id` parameter threaded from its one caller (`StartChatSessionUseCase`), since it never took a `usage_context` before (it doesn't call the model).

## Test plan

- [x] New unit test on the helper itself: under-cap/at-cap/over-cap boundaries, and log content (chapter_id + dropped_chars) when truncation fires.
- [x] New API-level tests for digest, quiz, and chat confirming truncation actually happens end-to-end (not just at the helper level).
- [x] Full backend suite passes (892 tests).
- [x] pyright clean, import-linter contracts hold, duplication check under threshold.
- [x] Proved the new tests can fail: temporarily disabled truncation, confirmed all 5 new tests failed, reverted.

Fixes #613

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/backend/src/application/learning/commands/chat/start_chat_session_use_case.py
+++ b/backend/src/application/learning/commands/chat/start_chat_session_use_case.py
@@ -83,7 +83,9 @@ class StartChatSessionUseCase:
 
         # 4. Seed the message history with the chapter content (no AI round-trip);
         #    the model reads it when the reader sends their first message.
-        session.message_history = self.ai_chat_service.seed_chat_context(content, CHAT_OPENER)
+        session.message_history = self.ai_chat_service.seed_chat_context(
+            content, CHAT_OPENER, chapter_id=chapter_id
+        )
         saved_session = await self.ai_chat_session_repo.create(session)
 
         logger.info(
--- a/backend/src/application/learning/protocols/ai_chat_service.py
+++ b/backend/src/application/learning/protocols/ai_chat_service.py
@@ -6,7 +6,7 @@ from src.domain.common.types import SerializedMessageHistory
 
 class AIChatServiceProtocol(Protocol):
     def seed_chat_context(
-        self, chapter_content: str, assistant_opener: str
+        self, chapter_content: str, assistant_opener: str, *, chapter_id: int
     ) -> SerializedMessageHistory: ...
 
     async def continue_chat(
--- a/backend/src/infrastructure/ai/ai_service.py
+++ b/backend/src/infrastructure/ai/ai_service.py
@@ -25,15 +25,19 @@ from src.infrastructure.ai.ai_agents import (
 
 logger = structlog.get_logger(__name__)
 
-#: Longest chapter text sent to the model as context, in characters.
-#:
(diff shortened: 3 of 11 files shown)


Tumetsu:
@jdbway Thank you for taking a look into this :+1:  Refactored the test suite a bit to not use as much patching since I'm trying to avoid it when I can and instead rely on DI.

jdbway:
You're the boss! Thanks again for allowing me to contribute, and thanks for catching my mistake of implementing a new note icon instead of re-using your existing icon. That was a detail oversight on my part.

PR #625, opened 2026-08-24
Fix nested dialogs clobbering scroll lock and leaking arrow-key paging

### LINKED ISSUE #620: Arrow keys page the dialog underneath when a nested dialog has no navigation
## What happens

With a note dialog open **on top of** a highlight or chapter dialog, pressing <kbd>←</kbd> / <kbd>→</kbd> pages the dialog *underneath*. The nested note dialog stays put showing the old note, and closing it reveals a different highlight than the one you started from.

## Repro

1. Open a book → **Highlights**
2. Open any highlight that has a linked note, → **Notes** tab → click the note (or, on `jdbway/flashcard-linked-note` / #619: **Flashcards** tab → **View note**)
3. With the note dialog open on top, press <kbd>←</kbd> or <kbd>→</kbd>
4. The highlight behind the note dialog navigates to the previous/next highlight

Same via `ChapterDetailDialog` on the Structure page.

## Root cause

`useDialogHorizontalNavigation` registers a dialog on the module-level `activeNavigationStack` only when it *has* navigation:

```ts
// frontend/src/components/dialogs/useDialogHorizontalNavigation.ts:63-70
useEffect(() => {
  if (!open || !hasNavigation) return;
  ...
  activeNavigationStack.push(id);
```

where `hasNavigation = totalCount > 1 && onNavigate`.

A nested `NoteViewDialog` is mounted without `onNavigate`, so it never pushes itself onto the stack. The parent dialog therefore remains `activeNavigationStack[last]`, and its `window` `keydown` listener keeps firing while the note dialog is the topmost modal.

`data-prevent-navigation` does not save it: the attribute sits only on the tab strip (`DialogTabs.tsx:49`), not the tab panel, and the nested dialog renders into its own portal at `document.body`, so `e.target.closest(...)` never reaches the parent dialog's markup.

## Where it is reachable from

- `LinkedNotesSection.tsx:115` — `<NoteDialogs>` inside the highlight and chapter dialogs' Notes tab (pre-existing)
- `FlashcardListCard.tsx:119` — the "View note" button added in #619

## Suggested fix

Move the stack registration off `hasNavigation` and onto `open` alone, and gate only the key *handling* on `hasNavigation`. A modal with nothing to page between should still block the modal below it from stealing arrow keys.

Both effects in the hook currently share the `!open || !hasNavigation` guard; the registration effect wants `!open` only.

## Test

The vitest-browser + msw harness already drives these dialogs (`NotesPage.test.tsx`, `StructurePage.test.tsx`), so this is cheap to pin: open a nested note dialog over a highlight dialog, press <kbd>→</kbd>, assert the underlying highlight did not change.

### LINKED ISSUE #621: Nesting a CommonDialog clears the parent's body scroll lock and scrolls the page to the top
## What happens

Open a dialog, open a second dialog on top of it, close the second one. While the first dialog is **still open**, the page behind it jumps to the top and becomes scrollable again.

## Repro

1. Open a book → **Highlights**, scroll down a fair way
2. Open a highlight → **Notes** tab → click a linked note (a second dialog opens on top)
3. Close the note dialog with its × button
4. The highlight dialog is still open, but the background has scrolled to the top and now scrolls freely behind the modal

Also reachable from `LinkedNotesSection` inside `ChapterDetailDialog`, and — once #619 lands — from the flashcard list card's "View note" button and the flashcard editor's linked-note preview.

## Root cause

`CommonDialog` locks the body itself rather than relying on MUI (`disableScrollLock={true}` at `CommonDialog.tsx:112`):

```ts
// frontend/src/components/dialogs/CommonDialog.tsx:80-102
useEffect(() => {
  if (!open) return;
  const scrollY = window.scrollY;

  document.body.style.overflow = 'hidden';
  document.body.style.position = 'fixed';
  document.body.style.top = `-${scrollY}px`;
  document.body.style.width = '100%';

  return () => {
    document.body.style.overflow = '';
    document.body.style.position = '';
    document.body.style.top = '';
    document.body.style.width = '';
    window.scrollTo(0, scrollY);
  };
}, [open]);
```

The effect assumes it is the only dialog. Two `CommonDialog`s open at once break both halves of that assumption:

- **On mount of the inner dialog**, `window.scrollY` already reads `0` — the body is `position: fixed`, so the document has no scroll. The inner dialog therefore overwrites the outer dialog's `top: -Spx` with `top: 0px`, discarding the saved offset.
- **On unmount of the inner dialog**, the cleanup clears *every* body style and calls `window.scrollTo(0, 0)`, even though the outer dialog is still open. The outer dialog's effect does not re-run (its `open` never changed), so its lock is gone for good.

`ConfirmationDialog` is exempt — it uses a raw MUI `Dialog` with no scroll-lock effect of its own.

## Suggested fix

Make the lock reference-counted and owned by one module rather than by each dialog instance:

- A module-level counter (or a `useScrollLock()` hook). The first locker records `window.scrollY` and applies the body styles; nested lockers only increment. The last unlocker restores the styles and the recorded offset.
- Store the saved offset alongside the counter, so a nested dialog can never overwrite it with `0`.

Alternative: drop the hand-rolled lock and set `disableScrollLock={false}`, letting MUI's `ModalManager` handle nesting — it already reference-counts. The custom lock exists for iOS Safari address-bar behaviour, so this needs checking on a real device before switching.

## Test

`CommonDialog.test.tsx` already exists. Add a case that mounts a nested `CommonDialog`, unmounts it, and asserts `document.body.style.position` is still `fixed` and `top` still holds the original offset while the outer dialog remains open.

## Related

Same class as #620 (nested dialogs and the assumptions single-dialog code makes), but an independent root cause.

jdbway:
## Summary

Two related bugs in the nested-dialog family (both filed as the same class of issue):

- **#621** — Nesting a second `CommonDialog` on top of an already-open one clobbered the outer dialog's body scroll lock. The inner dialog's mount effect read `window.scrollY` as `0` (the body was already `position: fixed` from the outer lock), and its unmount effect then cleared the body styles the still-open outer dialog needed — snapping the page to the top and unlocking scroll while the outer dialog was still visible.
- **#620** — `useDialogHorizontalNavigation` only registered a dialog on the active-navigation stack when it had paging of its own. A dialog opened on top of a paged one but without paging itself (a note viewed standalone, the "Add note" editor, the "Link existing note" picker) never registered, so the paged dialog underneath stayed "topmost" and kept handling arrow keys — paging content the user could no longer see.

## Changes

- `CommonDialog`'s scroll lock is now reference-counted at module level: only the first locker records scroll position/applies styles, only the last unlocker restores them.
- `useDialogHorizontalNavigation` now registers every open dialog on the stack regardless of whether it has navigation, gating only key *handling* on `hasNavigation`.
- `NoteEditorDialog` and `NotePickerDialog` now call the hook (with no paging props) purely to participate in the stack — they were never in scope for the original navigation feature, so they simply never registered at all. This gap wasn't in the original issue text; it surfaced through manual testing on a real deployment (Add note / Link existing note both still leaked arrow keys after the core hook fix).

## Note for reviewers

The navigation-stack registration is opt-in per dialog — a future nested dialog that skips the `useDialogHorizontalNavigation` call will silently reintroduce #620. Worth hardening later by moving stack registration into `CommonDialog` itself, alongside the scroll lock, rather than requiring every consumer to remember it. Left out of this PR to keep the fix scoped to the reported bugs.

## Test plan

- [x] New regression test in `CommonDialog.test.tsx`: mounting and unmounting a nested `CommonDialog` leaves the outer dialog's body scroll lock intact. Verified it fails against the pre-fix code and passes after.
- [x] New `useDialogHorizontalNavigation.test.tsx`: a nested dialog with no navigation blocks arrow keys from reaching the dialog beneath it, and closing it restores paging. Verified fails pre-fix, passes post-fix.
- [x] Full frontend suite (`npx vitest run`), lint, and `tsc --noEmit` all pass.
- [x] Manually verified on a real deployment with production-like data: Add note, Link existing note, and clicking an existing linked note all now preserve scroll position and no longer leak arrow-key paging, from both the Highlights and Flashcards nesting paths.

Fixes #621
Fixes #620

https://claude.ai/code/session_019qFihShtX9DpKmFEwhxqNU


Tumetsu:
@jdbway Thanks for fixing this. I took liberty to to implement the suggestion mentioned in the Note for reviewers section since it makes sense to have this functionality in the common dialog.

jdbway:
> @jdbway Thanks for fixing this. I took liberty to to implement the suggestion mentioned in the Note for reviewers section since it makes sense to have this functionality in the common dialog.

@temetsu fantastic! I love seeing the best, most future-proof solutions get implemented. I have some free time, let me know if there are any other issues in particular you'd like me to look at, otherwise I'll continue reviewing the issues for things that make sense for me to address given my skillset. Obviously I'm not a dev but I do have a bunch of experience in software and systems integrations, plus I'm happy to burn some Claude tokens from my subscription on this.

Tumetsu:
@jdbway Thanks! There are couple rather simple tasks I haven't gotten around to do. On phone app I find it difficult to refer to issues directly but there is an issue #626 about changing settings page texts.  Another simple one is #637 

Those would be helpful to get done, simple enough for Claude and easily testable manually.

Btw, when prompting ask claude to not write long comments.

jdbway:
I will take a look at those. Also, thank you for the advice on shortening up the comments. I did notice Claude trying to tell you how your own code works the other day and I told it not to do that. Please give me all the feedback - over time I will get the process refined and the communication succinct. See, your software is enhancing my vocabulary already!

Issue #415, opened 2026-07-19
Proactive prompts for writing chapter gist notes


Tumetsu:
## Motivation

Follow-up to #413 (user-written chapter gist notes). This is not a new artifact — it is the interaction-design question of how a gist comes into existence: **pull** (a "write your gist" action in the chapter dialog that fires when the user remembers to press it) vs. **push** (the app notices a well-chosen moment and invites the gist).

The tension, in both directions:

- **Pull-only habits die quietly.** Writing gists is exactly the kind of virtuous-but-effortful practice that stops in the second week unless something surfaces it at the right moment. If the gist habit depends on remembering a toolbar button, structure-view gist coverage will stay sparse.
- **Push done badly becomes nagging.** A prompt that feels like a chore poisons the activity it promotes — the user starts writing token gists just to dismiss the prompt, which defeats the entire "formulation is the understanding" purpose of #413. The line between an invitation and a Duolingo owl is thin.

## Timing wrinkle (worth evaluating deliberately)

Writing a gist **from memory** some time after finishing a chapter is itself retrieval practice. So prompt timing is pedagogical, not just UX:

- Prompting at highlight-sync time (typically hours after the reading session) forces a small memory reach — likely better for retention than prompting immediately after reading.
- Prompting immediately (if that moment were available) gives freshness and the highest completion rate, but less retrieval benefit.

## Candidate trigger designs

1. **Sync-inferred completion:** when a highlight sync brings in highlights for chapter N+1, that is a strong signal chapter N is finished → the book page (or the future "next up" card from #414) shows a small inline invitation: "You seem to have finished <chapter> — one sentence: what was it about?" with an inline input.
2. **Empty-slot affordance (softest push):** when opening `ChapterDetailDialog` for a chapter that has highlights but no gist, render a visible empty gist slot ("Your gist: — add") instead of burying the action in the toolbar. No interruption, but the gap is visible.
3. Possible escalation policy: never more than one pending invitation at a time; dismissing an invitation for a chapter permanently silences it for that chapter.

## Recommendation from the original discussion

Ship #413 with the manual (pull) entry point first and live with it for a while; decide on push triggers based on whether gist coverage actually stays sparse. This ticket exists so that evaluation is remembered and deliberate rather than the prompts being bundled reflexively into the first implementation.

## Open questions

- Which trigger(s), if any: sync-inferred completion, empty-slot affordance, or both?
[...]

PR #658, opened 2026-08-27
Write and edit a chapter gist inline

### LINKED ISSUE #635: Write a chapter gist inline, without opening the note dialog
## Problem

Writing a chapter gist costs five interactions. In `ChapterDetailDialog`, the Gist section shows a "Write gist" button; pressing it opens `NoteEditorDialog` — a second, nested dialog with Title, Kind, Note (markdown), Chapters and Tags fields — where the reader types one sentence, presses Save, and the dialog closes.

Every field but the body is already decided by the caller: `ChapterGistSection` passes `initialKind="gist"`, `initialChapterIds={[chapterId]}` and `initialTitle={chapterName}`. The reader confirms four pre-filled fields to write one line.

#413 shipped the gist; #415 argues the habit dies unless the app makes it easy at the right moment. A full modal for a single sentence is exactly the friction that kills it. Editing an existing gist costs the same five interactions.

## Scope

Write and edit a chapter gist inline in `ChapterDetailDialog`, without opening the note dialog. The note dialog stays available for the cases that need it (adding tags, linking extra chapters, longer markdown).

## What it touches

| File | Role |
| --- | --- |
| `frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx` | the whole change lives here |
| `frontend/src/pages/BookPage/Notes/NoteEditorForm.tsx` | source of the payload shape and the `GIST_LENGTH_GUIDE = 200` counter and helper text |
| `useCreateNote` / `useUpdateNote` (`@/api/generated/notes/notes.ts`) | already the mutations the form calls |
| `useCacheEvents().noteChanged(bookId, noteId)` | the invalidation the form does after saving |

The create payload is fully determined without the dialog:

```ts
{ book_id, title: chapterName, body, kind: 'gist',
  chapter_ids: [chapterId], highlight_ids: [], tag_ids: [] }
```

Note the API takes `title` and `NoteEditorForm` gates saving on a non-empty title (`canSave`). Inline, the title is the chapter name and the reader never sees it — the field is not a blocker, but the inline path must supply it.

Nothing on the backend changes.

## Design questions

- **Save trigger.** Blur-to-save is the fewest interactions and is the pattern used elsewhere here. Debounced autosave was deliberately removed for chapter question fields in #259 — do not reintroduce it here without reading that first.
- **Edit affordance.** Today a saved gist renders as italic text with a pencil. Inline editing could make the text itself click-to-edit, which also delivers #415's "empty-slot affordance" idea for free: render an always-visible empty input instead of a "Write gist" button.
- **Multiline.** A gist is 1–2 sentences with a 200-character guide. A single-line input enforces the spirit; a 2-row `multiline` field is friendlier when the reader overshoots. Enter should probably save rather than insert a newline.
- **Errors and pending state.** `NoteEditorDialog` shows "Saving…" in its footer and surfaces failures via `useMutationErrorHandler`. Inline needs its own equivalent — an inline error that keeps the reader's text is the minimum; losing typed text on a failed save is the one unacceptable outcome.

## Acceptance criteria

- A chapter with no gist can get one from inside `ChapterDetailDialog` without a second dialog opening.
- An existing gist can be edited in place and saved.
- A saved gist appears immediately on the Structure page's accordion rows, which read gists from `useGetNotesForBook(book.id, { kind: 'gist' })` (`StructurePage.tsx:70`).
- The created note is indistinguishable from one written through the note dialog: `kind: 'gist'`, linked to the chapter, titled with the chapter name.
- A failed save keeps the reader's text and says so.
- The 200-character guide from `NoteEditorForm` is preserved in some form.

## Related

- #413 — user-written chapter gists (shipped; this is its interaction follow-up)
- #415 — proactive prompts for writing gists; its "empty-slot affordance" option is the same UI as this
- #259 — debounce autosave was removed from chapter question fields, for reasons that apply here

### LINKED ISSUE #660: Share the save-on-blur field logic between the gist and digest-answer fields
## Problem

Three fields in the app carry the same "local draft over a server value, committed on blur" logic, each with its own copy:

| Site | What it has |
| --- | --- |
| `ChapterGistSection` (#635) | draft state, Enter-blurs-to-save, Escape revert, a `cancelled` ref so Escape's blur does not save, failure keeps the text |
| `ChapterReviewSection` digest answers | `localEdits` keyed by chapterId, blur saves all answers — no Escape, no failure handling at all |
| `TagGroupHeader` rename | Enter saves, Escape cancels, `ClickAwayListener` + `onBlur` |

The fiddly part is not the blur wiring. It is the rule for **when the local draft yields to the server value**: drop it too early and the old text flashes back between save and refetch; never drop it and a background refetch is invisible. Getting it wrong is how #259 happened — a late response overwriting newer keystrokes. That rule deserves one implementation.

## Proposal

A headless hook, not a HoC — a HoC would have to own the input, and the call sites want different MUI props (multiline, counters, `error`, `autoFocus`).

```ts
// src/hooks/useCommitOnBlur.ts
const field = useCommitOnBlur({
  saved: gist?.body ?? '',            // the server's value
  onCommit: (value) => save(value),   // fires once, only when the value changed
  submitOnEnter: true,                // Shift+Enter always inserts a newline
});
// → field.value, field.isDirty, field.revert()
// → field.inputProps: { value, onChange, onBlur, onKeyDown }
```

**The hook owns**

1. Draft over server value (`draft ?? saved`).
2. Dropping the draft once the refetch carries it (`useResetOnChange([saved])`).
3. Committing exactly once — Enter blurs rather than saving, so Enter-then-click-away stays one request.
4. Escape: revert, suppress the blur it causes, and `stopPropagation` so a dialog does not close instead.
5. No commit when nothing changed.

**The call site keeps** which mutation runs, what an empty value means (a gist deletes, a digest answer legitimately saves empty), any prose ↔ field toggle, and how failure is shown.

## Scope

- `src/hooks/useCommitOnBlur.ts` plus unit-level coverage of the draft/refetch rule.
- Migrate `ChapterGistSection` — roughly 35 lines of plumbing down to 8.
- Migrate the digest answers, which needs a small `DigestAnswerField` component so the hook is not called in a loop. **This is a behaviour change**: those fields gain Escape and gain failure reporting they do not have today, so they need route-level tests of their own.
- `TagGroupHeader` fits too, lowest value — optional.
- `SearchBar` is out: it commits a search, not a save, and has no server value to fall back to.

## Why not in #635

An abstraction extracted from a single call site is a guess. Migrating the digest answers is what proves the seam sits in the right place, and it touches a different feature than the PR it would ride along in.

Follow-up to #635 — review discussion: https://github.com/Crossbill-App/crossbill-web/pull/658

Tumetsu:
Fixes #635
Fixes #660

## What

Writing a gist cost five interactions: a "Write gist" button in `ChapterDetailDialog` opened `NoteEditorDialog`, whose Title, Kind and Chapters fields the caller had already filled in. The reader confirmed four pre-filled fields to write one line.

The gist now lives in the chapter dialog itself. A chapter without one shows the input; a chapter with one shows the prose until it is clicked.

- **Blur and Enter save.** Enter blurs the field rather than saving directly, so the save happens once whether the reader finishes with Enter or by clicking away. Shift+Enter inserts a newline.
- **Escape reverts** the text and closes the field, and stops propagating so it does not close the chapter dialog underneath.
- **No debounced autosave.** #259 removed it from the chapter question fields because an in-flight response overwrote later keystrokes; the same race applies here.
- **A failed save keeps the text**, marks the field and says "Not saved — try again." next to the character counter.
- **An update carries over** the note's title, kind, tags and links, so a gist given extra chapters or tags in the note dialog keeps them.

## Two decisions beyond the ticket

- **Clearing the text deletes the note**, with a "Gist deleted." snackbar, no confirmation and no undo — `SnackbarContext` has no action slot. It deletes even when the note carries tags or links to other chapters. Recorded on the ticket: https://github.com/Crossbill-App/crossbill-web/issues/635#issuecomment-5444816249
- **The section is no longer collapsible, and no longer routes to `NoteEditorDialog`.** Tags and extra chapter links on a gist are reachable from the Notes page only.

This also ships #415's "empty-slot affordance" option — the empty chapter renders the input rather than a button. #415 is closed as not planned.

## Structure

`GIST_LENGTH_GUIDE` and its helper row moved out of `NoteEditorForm` into `GistHelperText`, so the inline field and the note dialog show the same 200-character guide (first commit, separately reviewable).

## Shared save-on-blur field (#660)

Added after review. `useCommitOnBlur` (`src/hooks/useCommitOnBlur.ts`) owns the part that is identical wherever a field saves on blur, and is the only place the #259 race is guarded:

- the draft-over-server merge, with the draft yielding only once the refetch carries it **and** the field is unfocused
- committing exactly once — Enter blurs rather than saving, and leaving an unchanged field twice before the response lands does not save twice
- Escape revert, with the blur it causes suppressed and `stopPropagation` so a hosting dialog does not close instead

The call sites keep which mutation runs, what an empty value means, and how failure is shown.

**The digest answer fields gain** Escape, a reported failure instead of a silent one, and protection from the double save. Their `localEdits` map is gone — the hook holds the text until the save round-trips — and the payload now patches one question index instead of replacing every answer, which matches `ChapterDigest.update_user_answers` on the backend.

`TagGroupHeader`'s rename fits the hook too, but it has no test coverage today, so migrating it stays out of this PR — #660 lists it as optional.

New tests: 4 on the hook (`useCommitOnBlur.test.tsx`) for the draft/refetch rule, the double-commit guard, and Enter/Escape; 4 route-level on the digest answers (`ChapterReviewSection.test.tsx`). Each was checked to fail when the behaviour it covers is broken.

## How to verify

- `npm run test` — 108 passed, including 5 tests in `ChapterGistSection.test.tsx`: create without a second dialog opening, edit in place preserving tags and chapter links, Escape revert, clear-to-delete, and failed-save-keeps-text.
- Each new test was checked to fail when the behaviour it covers is broken.
- `lint`, `type-check`, `knip` and `duplication-check` pass. Duplication is 1.3% before and after, so the 1.33 threshold is unchanged.
- `tests/msw/bookApi.ts` gained `POST /api/v1/notes` and `DELETE /api/v1/notes/:noteId`, both stateful, so the assertions are round-trips rather than optimistic UI.

## Not done

Nothing on the backend changes. Gist deletion is still not offered anywhere except by emptying the field — `NoteEditorDialog` has no delete, and that was left alone.

https://claude.ai/code/session_01Ft2M6DWiK3dVYSDCGzt5Mv

### DIFF
--- a/frontend/src/hooks/useCommitOnBlur.ts
+++ b/frontend/src/hooks/useCommitOnBlur.ts
@@ -1,0 +1,120 @@
+import { useRef, useState, type ChangeEvent, type KeyboardEvent } from 'react';
+import { useResetOnChange } from './useResetOnChange';
+
+interface UseCommitOnBlurOptions {
+  /** The value the server holds. */
+  saved: string;
+  /** Called when the field is left carrying a value the server does not have. */
+  onCommit: (value: string) => void;
+  /** Called when the field is left, changed or not. */
+  onBlur?: () => void;
+  /** Called when Escape reverts the field. */
+  onCancel?: () => void;
+  /** Enter commits, Shift+Enter inserts a newline. Off for free-form prose. */
+  submitOnEnter?: boolean;
+}
+
+export interface CommitOnBlurField {
+  value: string;
+  isDirty: boolean;
+  /** Drop the local text and show `saved` again. */
+  revert: () => void;
+  /** After a failed save: let leaving the field commit the same value again. */
+  allowRecommit: () => void;
+  inputProps: {
+    value: string;
+    onChange: (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+    onFocus: () => void;
+    onBlur: () => void;
+    onKeyDown: (event: KeyboardEvent<HTMLDivElement>) => void;
+  };
+}
+
+/**
+ * A text field whose local text is committed when the reader leaves it.
+ *
+ * The caller keeps everything domain-shaped — which mutation runs, what an
+ * empty value means, how a failure is shown.
+ */
+export const useCommitOnBlur = ({
+  saved,
+  onCommit,
+  onBlur,
+  onCancel,
+  submitOnEnter = true,
+}: UseCommitOnBlurOptions): CommitOnBlurField => {
+  // The text on screen while a save is in flight; `null` means the server's
+  // value is what is shown.
+  const [draft, setDraft] = useState<string | null>(null);
+  const isFocused = useRef(false);
+  // The value last handed to `onCommit`, so leaving the field again before the
+  // refetch lands does not save it twice — which for a create would mean two
+  // records rather than one.
+  const committed = useRef<string | null>(null);
+  // Set by Escape, so the blur it causes does not commit the reverted text.
+  const cancelled = useRef(false);
+
+  const revert = () => {
+    setDraft(null);
+    committed.current = null;
+  };
+
+  // While the field has focus the reader's text wins. Once they have left, the
+  // refetched value takes over — dropping the draft any earlier would flash the
+  // old text back between the save and the refetch.
+  useResetOnChange([saved], () => {
+    if (!isFocused.current) revert();
+  });
+
+  const value = draft ?? saved;
+
+  const commit = () => {
+    if (value === saved || value === committed.current) return;
+    committed.current = value;
+    onCommit(value);
+  };
+
+  return {
+    value,
+    isDirty: value !== saved,
+    revert,
+    allowRecommit: () => {
+      committed.current = null;
+    },
+    inputProps: {
+      value,
+      onChange: (event) => {
+        cancelled.current = false;
+        committed.current = null;
+        setDraft(event.target.value);
+      },
+      onFocus: () => {
+        isFocused.current = true;
+      },
+      onBlur: () => {
+        isFocused.current = false;
+        if (cancelled.current) {
+          cancelled.current = false;
+        } else {
+          commit();
+        }
+        onBlur?.();
+      },
+      onKeyDown: (event) => {
+        if (submitOnEnter && event.key === 'Enter' && !event.shiftKey) {
+          // Blur rather than commit here, so the value is saved once whether
+          // the reader finishes with Enter or by clicking away.
+          event.preventDefault();
+          (event.target as HTMLElement).blur();
+        } else if (event.key === 'Escape') {
+          // A dialog hosting the field closes on Escape; reverting the text is
+          // what this Escape means.
+          event.stopPropagation();
+          cancelled.current = true;
+          revert();
+          onCancel?.();
+        }
+      },
+    },
+  };
+};
--- a/frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
+++ b/frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
@@ -12,6 +12,30 @@ import { useRef, useState, type KeyboardEvent } from 'react';
 
 const PLACEHOLDER = 'What was this chapter about?';
 
+interface ExistingGistTextProps {
+  text: string;
+  onEdit: () => void;
+}
+
+const ExistingGistText = ({ text, onEdit }: ExistingGistTextProps) => (
+  <ButtonBase
+    onClick={onEdit}
+    sx={{
+      width: '100%',
+      justifyContent: 'flex-start',
+      textAlign: 'left',
+      borderRadius: 1,
+      px: 1,
+      py: 0.5,
+      '&:hover': { bgcolor: 'action.hover' },
+    }}
+  >
+    <Typography sx={{ fontStyle: 'italic' }} variant="body1">
+      {text}
+    </Typography>
+  </ButtonBase>
+);
+
 interface ChapterGistSectionProps {
   chapterId: number;
   chapterName: string;
@@ -145,22 +169,7 @@ export const ChapterGistSection = ({ chapterId, chapterName, notes }: ChapterGis
         Gist
       </Typography>
       {gist && !isEditing ? (
-        <ButtonBase
-          onClick={() => setIsEditing(true)}
-          sx={{
-            width: '100%',
-            justifyContent: 'flex-start',
-            textAlign: 'left',
-            borderRadius: 1,
-            px: 1,
-            py: 0.5,
-            '&:hover': { bgcolor: 'action.hover' },
-          }}
-        >
-          <Typography sx={{ fontStyle: 'italic' }} variant="body1">
-            {value}
-          </Typography>
-        </ButtonBase>
+        <ExistingGistText text={value} onEdit={() => setIsEditing(true)} />
       ) : (
         <TextField
           fullWidth
--- a/frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
+++ b/frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
@@ -1,11 +1,16 @@
-import type { NoteWithLinks } from '@/api/generated/model';
-import { IconButtonWithTooltip } from '@/components/buttons/IconButtonWithTooltip.tsx';
-import { NoteEditorDialog } from '@/pages/BookPage/Notes/NoteEditorDialog';
-import { EditIcon } from '@/theme/Icons.tsx';
-import { Box, Button, Typography } from '@mui/material';
+import type { NoteUpdateRequestKind, NoteWithLinks } from '@/api/generated/model';
+import { useCreateNote, useDeleteNote, useUpdateNote } from '@/api/generated/notes/notes.ts';
+import { useSnackbar } from '@/context/SnackbarContext.tsx';
+import { useMutationErrorHandler } from '@/hooks/useMutationErrorHandler.ts';
+import { useResetOnChange } from '@/hooks/useResetOnChange.ts';
+import { useCacheEvents } from '@/lib/cacheEvents.ts';
+import { useBookPage } from '@/pages/BookPage/BookPageContext';
+import { GistHelperText } from '@/pages/BookPage/Notes/GistHelperText.tsx';
+import { Box, ButtonBase, TextField, Typography } from '@mui/material';
 import { find } from 'lodash';
-import { useState } from 'react';
-import { CollapsibleSection } from './CollapsibleSection';
+import { useRef, useState, type KeyboardEvent } from 'react';
+
+const PLACEHOLDER = 'One sentence: what was this chapter about?';
 
 interface ChapterGistSectionProps {
   chapterId: number;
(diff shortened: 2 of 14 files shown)


frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
```diff
@@ -1,57 +1,196 @@
...
-            <IconButtonWithTooltip
-              title="Edit gist"
-              ariaLabel="Edit gist"
-              icon={<EditIcon fontSize="small" />}
-              onClick={() => setEditorOpen(true)}
+    <Box sx={{ px: 2, mb: 2 }}>
+      <Typography variant="body1" sx={{ fontWeight: 600, color: 'primary.main', py: 1.5 }}>
+        Gist
+      </Typography>
+      {gist && !isEditing ? (
```
Tumetsu, line 141: Move the contents of this section into some well named subcomponent in this file. e.g. "ExistingGistText" etc.

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
```diff
@@ -1,57 +1,196 @@
...
+  const cancelled = useRef(false);
+
+  useResetOnChange([chapterId], () => {
+    setDraft(null);
+    setIsEditing(false);
+    setSaveFailed(false);
+  });
+
+  // Once the refetched note carries the draft, the server value takes over —
+  // dropping the draft any earlier would flash the old body on screen.
```
Tumetsu: Looking at this implementation for the field here with onBlur etc. handlers I wonder could we make a common SaveOnBlurInput etc. hook or HoC which could be reused across the app? The question answers in the Chapter digest dialog for example could use the same implementation. Ofc the handlers and triggered mutations would change but something like this might be nice. 

Think and give me a suggestion about possible implementation before acting on it.

Tumetsu: Agreed on the shape, but proposing it as a follow-up rather than here: #660.

Short version — a headless hook, `useCommitOnBlur({ saved, onCommit, submitOnEnter })` returning `inputProps` to spread onto whatever `TextField` the site already renders. A HoC would have to own the input, and the call sites want different MUI props (multiline, counter, `error`, `autoFocus`).

The hook owns the draft-over-server merge, dropping the draft once the refetch carries it, committing exactly once (Enter blurs rather than saving), Escape revert with the blur it causes suppressed, and no-commit-when-unchanged. The call site keeps which mutation runs, what an empty value means, and how failure is shown.

Why not in this PR: an abstraction extracted from one call site is a guess — migrating the digest answers is what proves the seam. That migration is also a behaviour change, since those fields gain Escape and failure reporting they don't have today, so it wants its own tests and review.

The full breakdown, including why `SearchBar` is out, is in #660.

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
Tumetsu: Done in d22786b0 — the prose branch is now an `ExistingGistText` component at the top of the file, taking `text` and `onEdit`.

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
Tumetsu: Implemented here after all, per your call — `aa11010c`.

`useCommitOnBlur` lives in `src/hooks/useCommitOnBlur.ts` and both sites use it. The digest answers came out better than expected: their chapter-keyed `localEdits` map is gone entirely, because the hook already holds the text until the save round-trips. They also gained Escape, a reported failure instead of a silent one, and protection from a double save when the field is left twice before the response lands.

One thing the migration turned up: the answers payload was replacing every answer on every save, but `ChapterDigest.update_user_answers` patches by question index — so it now sends just the one that changed.

`TagGroupHeader` is left alone: it fits the hook, but it has no test coverage today, so an untested refactor of a different feature seemed the wrong thing to bundle in. #660 is moved to In review and lists it as optional.

frontend/src/hooks/useCommitOnBlur.ts
```diff
@@ -0,0 +1,125 @@
...
+    onFocus: () => void;
+    onBlur: () => void;
+    onKeyDown: (event: KeyboardEvent<HTMLDivElement>) => void;
+  };
+}
+
+/**
+ * A text field whose local text is committed when the reader leaves it.
+ *
+ * Deliberately not debounced: #259 removed a debounced autosave because the
```
Tumetsu: Remove this #259 reference from the comment and the explanation why this is not debounced.

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
```diff
@@ -1,57 +1,173 @@
...
+
+const PLACEHOLDER = 'What was this chapter about?';
+
+interface ExistingGistTextProps {
+  text: string;
+  onEdit: () => void;
+}
+
+/**
+ * A written gist, as prose. A `ButtonBase` rather than a click handler on the
```
Tumetsu: Remove this comment block

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
```diff
@@ -1,57 +1,173 @@
...
+  </ButtonBase>
+);
 
 interface ChapterGistSectionProps {
   chapterId: number;
   chapterName: string;
   notes: NoteWithLinks[];
 }
 
+/**
```
Tumetsu: remove this comment block

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterReviewSection.tsx
```diff
@@ -12,12 +12,48 @@ import { AIActionButton } from '@/components/buttons/AIActionButton.tsx';
...
 import { CollapsibleSection } from './CollapsibleSection.tsx';
 
+interface DigestAnswerFieldProps {
+  question: string;
+  savedAnswer: string;
+  onSave: (answer: string, onError: () => void) => void;
+}
+
+/**
+ * One digest question and the reader's answer to it, saved when they leave the
```
Tumetsu: remove this comment block

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterReviewSection.tsx
```diff
@@ -37,21 +73,15 @@ export const ChapterReviewSection = ({
 }: ChapterReviewSectionProps) => {
   const queryClient = useQueryClient();
   const cache = useCacheEvents();
+  const mutationErrorHandler = useMutationErrorHandler();
 
-  // Local edits keyed by chapterId so they reset when switching chapters
-  const [localEdits, setLocalEdits] = useState<Record<number, Record<number, string>>>({});
-  // Server answers derived from digestSummary
-  const serverAnswers = useMemo<Record<number, string>>(() => {
+  // Each field holds its own text until the save round-trips, so there is no
```
Tumetsu: remove this comment

frontend/src/hooks/useCommitOnBlur.ts
Tumetsu: Removed — the docstring keeps only what the hook is and what the caller owns. Folded into `afeb6bda` (was `aa11010c`).

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
Tumetsu: Removed. Folded into `eba7addf` (was `d22786b0`).

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterGistSection.tsx
Tumetsu: Removed. Folded into `3b1e429f` (was `b4a423b3`).

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterReviewSection.tsx
Tumetsu: Removed. Folded into `afeb6bda` (was `aa11010c`).

frontend/src/pages/BookPage/Structure/ChapterDetailDialog/ChapterReviewSection.tsx
Tumetsu: Removed. Folded into `afeb6bda` (was `aa11010c`).

Tumetsu:
CI failure fixed — it was a defect in the test, not flakiness.

`await expect.element(dialog.getByText('Attention is selective.')).toBeVisible()` matched the field's own `<textarea>`, which already holds that text while the create is still pending. I verified this with a probe: at that point the matcher finds exactly one element, a `TEXTAREA`. So the assertion passed instantly without waiting for anything, and `expect(state.notes).toHaveLength(1)` on the next line raced the mocked POST resuming. Locally the microtasks won that race; on the CI runner they did not.

The assertion now waits for the saved gist's button, which only renders once the note exists in the refetched query data:

```ts
await expect
  .element(dialog.getByRole('button', { name: 'Attention is selective.' }))
  .toBeVisible();
```

The same weak assertion in "a gist is written from the chapter dialog without a second dialog opening" got the same treatment. It was not failing, because an `expect.poll` on the state followed it, but it was not proving the prose rendered either.

Both fixes are folded into the commits they belong to (`d62ff970`, `5802bfcf`). Checked that the pending-create test still fails when `disabled={createMutation.isPending}` is removed, so it still guards what you wrote it for.

PR #659, opened 2026-08-27
Add book blurb to the book header area (#653)

### LINKED ISSUE #653: Add book blurb to the book header area
Image

Should be taken from uploaded book if the metadata provides one or let user to edit it via simple text field.

It is slightly problematic in practice since text in the metadata might be strangely formatted. Perhaps MVP is to just add an editable field which renders markdown subset? The field could be located and editable in the manage book dialog which currently opens from the Edit button visible in the screenshot.

Tumetsu:
Closes #653.

A book's blurb now shows in the book-page header and can be written or edited in the Manage Book dialog.

`Book.description` already existed end to end — domain entity, ORM `Text` column, `BookDetails` schema, both read models — with two gaps: nothing rendered it, and nothing wrote it after the book was created. This fills both.

## What changed

**Backend** — `PATCH /api/v1/books/{book_id}`, returning 204.

- Partial by contract: the router reads `model_fields_set`, so a key the client omits is left untouched while an explicit `null` clears. The existing PATCH precedent (`highlight-labels`) conflates the two; that is safe on labels but not on books, where `Book.__post_init__` rejects an empty title.
- `extra="forbid"` on the request schema, so an unrecognised key 422s instead of silently returning 204 with nothing written. A deliberate local deviation — no other schema in `src/infrastructure` does this — because this endpoint's semantics are defined by which keys are present.
- Blank or whitespace-only text normalises to `null` in the domain setter, and the blurb is capped at 5000 characters.

**Frontend**

- `BookBlurb` renders under the header stats, clamped to three lines with a Show more / Show less toggle.
- Rendered through `react-markdown` with `rehype-raw` + `rehype-sanitize`. Publisher metadata arrives as HTML far more often than as Markdown, so without `rehype-raw` the tags show as literal text; that makes `rehype-sanitize` mandatory rather than optional, since the same field is now reader-writable.
- A multiline Blurb field and a Save button in `BookEditDialog`, gated on dirty state.

## Deliberately out of scope

- **Extracting the blurb from uploaded epub metadata.** `EpubParserProtocol` parses TOC, cover, and validity only, and existing epubs would need a backfill job. Worth a follow-up ticket — which must write only when `description is None`, so it cannot clobber a blurb the reader wrote.
- Editing title, author, or ISBN. `PATCH` is additive, so they can be added without a new endpoint. The first second field is the moment to introduce an `UpdateBookInput` DTO with an explicit `UNSET` sentinel rather than growing per-field branches in the router.
- Showing the blurb on the library list.

## Testing

Backend 919 passing, frontend 100 passing, pyright / ruff / type-check / lint / knip clean, jscpd 2.3% and 1.3% (both under threshold).

New coverage: the four PATCH semantic states (set, explicit null, blank-to-null, omitted), over-limit, unknown key, 404, and user isolation asserting the stranger's row is unchanged after the 404. On the frontend: the save round-trip reaching the header, clearing a blurb, the error path (snackbar reports it, the typed text survives, the original blurb stays on screen), and a sanitiser test asserting `<script>` and `img[onerror]` are stripped.

## Still worth a human eye

Open a book whose blurb came from the KOReader plugin's epub metadata and confirm it renders as formatted text rather than visible tags, and that a multi-paragraph blurb clamps cleanly — `-webkit-line-clamp` counts lines across block children, so a second paragraph's margin can show as a sliver.

https://claude.ai/code/session_01Kkc5PCZvrR5MbDNSYry6Tp


frontend/src/pages/BookPage/BookTitle/BookBlurb.tsx
```diff
@@ -0,0 +1,117 @@
...
+          aria-expanded={expanded}
+        >
+          {expanded ? 'Show less' : 'Show more'}
+        </Button>
+      )}
+    </Box>
+  );
+};
+
+/**
```
Tumetsu: Remove this comment block

frontend/src/pages/BookPage/BookTitle/BookEditDialog.tsx
```diff
@@ -1,14 +1,28 @@
...
 import { DeleteIcon } from '@/theme/Icons.tsx';
 import { Box, Button, Typography } from '@mui/material';
 import { useNavigate } from '@tanstack/react-router';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
+import { useForm } from 'react-hook-form';
+
+const BLURB_MAX_LENGTH = 5000;
+
+/**
```
Tumetsu: remove this comment block

frontend/src/pages/BookPage/BookTitle/BookEditDialog.tsx
```diff
@@ -44,27 +58,62 @@ export const BookEditDialog = ({ book, open, onClose }: BookEditDialogProps) =>
...
+  const {
+    control,
+    handleSubmit,
+    reset,
+    formState: { isDirty },
+  } = useForm<{ description: string }>({
+    defaultValues: { description: seedBlurb(book.description) },
+  });
+
+  // Re-seed on open only, so a closed dialog's abandoned draft never survives
```
Tumetsu: remove comments

frontend/src/pages/BookPage/BookTitle/BookTitle.tsx
```diff
@@ -88,14 +90,24 @@ export const BookTitle = ({ book }: BookTitleProps) => {
...
           }}
         >
-          <Typography
-            variant="h1"
-            component="h1"
-            sx={{
-              mb: 1,
-            }}
-          >
+          {/* The button flows as part of the heading's text so it follows the
```
Tumetsu: remove this comment

PR #663, opened 2026-08-28
UI audit: the agreed findings in sections A–J

Tumetsu:
Implements the agreed findings from the Crossbill UI Audit, sections A–J — one commit each.

## Section A

| Commit | Finding | Change |
|---|---|---|
| `bb44020b` | **A6** | Deleting a tag group went straight from the icon's `onClick`. It now routes through `ConfirmationDialog`, naming the group and saying its tags survive and move to Ungrouped. The tag-group action icons gained accessible names. |
| `bce95905` | **A9** | The note picker's three names become **Link a note** with the target as subtitle, from a union prop so the call sites cannot drift. |
| `e9d52457` | **A8** | A shared `CollapseChevron`; the hand-rolled chevrons route through it — the tag-group one rotated the other way. |
| `89be19d1` | **A4** | `NoteViewDialog`'s in-place editor is the one edit surface; the reflection page opens it with `initiallyEditing`. `NoteEditorDialog` is create-only. Footer Save/Cancel sit at the right edge. |
| `788311a9` | **A5** | One shared `SavedIndicator` on all six autosaving edits. |
| `f5f16b6f` | fix | The flashcard action cluster was absolutely positioned over a fixed reserve too narrow for three icons; laid out in the flow instead. |
| `6991f3d2` `b80ee690` | fix | The saved marker fades both ways, and its reserved space matches what it occupies so nothing shifts. |
| `84180088` | revert | The flashcard chevron is out again — the card wants a deeper redesign, leaving that half of A8 open. |

## Section B

| Commit | Finding | Change |
|---|---|---|
| `c8fbb061` | **B1** | Both colour declarations read a backtick string, so the browser dropped them and every highlight card rendered its metadata near-black. `text.secondary`. |
| `381252ef` | **B10** | The reading-stage marker's `'primary' \| 'secondary'` colour only ever emitted `'primary'` — a branch for a palette that does not exist. |
| `8f417c14` | **B4** | Three periods for an ellipsis, recorded in the frontend conventions. |
| `0f7c5db0` | **B11** | `ConfirmationDialog` is no longer full-screen below `sm`, where it covered the thing being asked about. |
| `cb6f7e48` | **B3** | One rendered date format behind both formatters and the pickers; the en-US pin stays for parsing. |
| `b420b629` | **B2** | The bullet-plus-four-nbsp separator becomes a middot, and the two hand-rolled rows go through `MetadataRow`. |
| `24ca8b06` | **B8** | `EmptyStateText` gains a `page` variant; all six hand-rolled empty states use it. |
| `76afd1e0` | **B5** | `PageTitle` on every page, with a `component` prop so heading levels stay right. |
| `49314dcb` | **B12** | The auth pages use the themed card instead of `Paper elevation={3}`. |
| `ca039fae` | **B7** | Chapters and bookmarks use `SidebarSectionHeader`, whose collapse toggle now lives on the button with `aria-expanded` — it previously had no handler and worked only by bubbling. |
| `be09ba7a` | **B6** | Every section heading renders through `SectionTitle`'s two variants. |
| `0367be6b` | **B9** | The card's left rail is a type marker only; hover is a tint and a lift on all three cards. |

Plus `b2575d27` (import order and indentation, formatting only) and `b68a19b4` (duplication ratchet, 1.25 → 1.1, after the shared components landed).

## Where I read a finding differently

- **B6** says to delete `SidebarSectionHeader`; **B7**, also agreed, says to extend it. Kept the component — B7's three callers need its icon, action slot and collapse control — and replaced its *heading text* with `SectionTitle`, which is what made it a fourth treatment.
- **A8** stops short of B7's consolidation in its own commit, and its flashcard half was reverted on review.
- **B7** also drops the chevron's `display: { xs: 'none', lg: 'block' }`: an affordance that hides while its action still works is the thing the finding is about.

## Section C

| Commit | Finding | Change |
|---|---|---|
| `4e78ebb8` | **C1** | Seven labels for committing a form become one rule: **Save** for editing something that exists, a named create (`Add flashcard`, `Add group`) only where the form is one of several controls on screen, and `Saving...` as the single pending word. Recorded in the frontend conventions. |
| `2e8e8bb8` | **C2** | The new-group form takes the dialog footer convention — right-aligned text Cancel beside a contained primary — instead of two half-width buttons. |
| `cf74556c` | **C3** | Every AI action goes through `AIActionButton`, which gains an `iconOnly` form so the toolbars keep their icon convention, and an `icon` override for regenerate. The library-wide embedding batch is no longer the only contained button. |
| `ff05cdb1` | **C5** | Scroll-to-top goes neutral so amber means one thing in that stack; the filter FAB keeps it and adds a count badge, taking a count rather than a boolean so its accessible name says how many filters are on. |
| `4249eaf4` | **C6** | One labelled `Add group` button in both the sidebar and the filter drawer. |

C4 was removed from the audit note before implementation, so it is not covered here.

## Spun out

[#664](https://github.com/Crossbill-App/crossbill-web/issues/664) — normalise KOReader highlight timestamps to ISO in the backend. B3 unified rendering, but the frontend still needs a bespoke locale-pinned parser because `highlight.datetime` is passed through in the device's own format.

## Section D

| Commit | Finding | Change |
|---|---|---|
| `8fc88141` | **D4** | The clear control on every search field was a `Box component="span"` holding a bare ✕ — not focusable, no accessible name, no keyboard activation. Now an `IconButton` with `aria-label="Clear search"`, keeping the `onMouseDown` focus guard. |
| `9e27e9a9` | **D7** | The two-item account menu gets `AccountCircle` instead of a hamburger; `MenuIcon` leaves the registry with it. |
| `a7cb0900` | **D1** | `HighlightsIcon` and `BookCoverIcon` were both `MenuBook`. Highlights take `FormatQuote`, already the highlight mark on cards and in the dialog, and `QuoteIcon` folds into `HighlightsIcon` — one glyph, one name. |
| `11183519` | **D3** | `Send`, `Replay` and the suggestion card's `Check`/`Close` move into the registry, with a `@typescript-eslint/no-restricted-imports` rule keeping it true — type imports still pass, both the named and deep-path icon forms are errors. |
| `86ae2dd5` | **D5** | A scale in `theme/iconSizes.ts` — 16 inline, 20 UI, 24 prominent — replacing six ad-hoc sizes. The tag-group actions go from 14px to 20px, since nothing clickable should be below `ui`. |

D2 and D6 are marked "needs discussion" in the audit note and are untouched. D3 stops at naming accept/reject rather than replacing the dismiss glyph, which is D2's question.

Both new rules — icons from the registry, sized from the scale — are recorded in the `frontend-design` skill.

## Section E

| Commit | Finding | Change |
|---|---|---|
| `3a5d31a1` | **E2** | Question was `primary.main` and Answer `secondary.main` under the same uppercase caption, a distinction carrying no meaning. `text.secondary` for both. |
| `7fdb89fd` | **E5** | The tag-group actions were `text.disabled` — the disabled colour on three enabled controls, one a delete — now `text.secondary`. The flashcard action cluster is dimmed only under `@media (hover: hover)`, since on touch nothing ever restores it. |
| `25c887ac` | **E4** | `whiteOverlay.group`/`.ungrouped` were translucent whites tuned to the stone page background, invisible on the paper-white filter drawer. Two opaque tokens from the stone ramp, under a new `surfaces` group; `whiteOverlay` keeps only the two that sit over the amber app bar. |
| `cb26e6d0` | **E1** | `filterChipSx(isSelected)` owns the padding, transition and hover lift for every chip that selects something. The label chip still overrides the colour, because a highlight label's colour is the one that means something. |

E3 is marked "needs discussion" and is untouched — it proposes reducing the saturated label chip to a dot, which overlaps E1's decision that a label keeps its own colour.

The audit's fifth "chip" in E1 is the filter FAB, which is a FAB rather than a chip; C5 settled what its amber means.

## Section F

| Commit | Finding | Change |
|---|---|---|
| `551e9acc` | **F1** | The generated chapter artifact had five names, one of them the backend's. "Digest" leaves the interface — `Generating digest...` becomes `Generating summary...`, `No digest questions available.` becomes `No questions generated for this chapter.` — and the dialog's first tab is **Questions** rather than "Chapter review". Three names remain: summary, questions, gist. The API, hooks, cache events and component names keep `digest`. |
| `283c4cfc` | **F2** | The editor said Kind, the sidebar Note types, the drawer tab Types. One word: **Type** on the select, **Types** on the filter section. |
| `cf38d536` | **F3** | `BOOK_PAGE_LABELS` holds each tab's name once; the desktop nav, the mobile nav, the overflow menu and each page's `PageTitle` all read it. "Structure of the book" and "Reading sessions" take the nav's shorter wording, and the per-route `label` field is gone. |
| `df988b57` | **F4** | "Recently Viewed" and "Recently read" are both sentence case, and the second becomes **Recently synced** — which is what `useGetRecentlySyncedBooks` actually returns, and the more useful claim. |
| `e8c8190d` | **F5** | Sentence case in sixteen strings: eight dialog and tooltip titles, five password labels, `Generate summary`, and the `Unknown author` / `Unknown chapter` / `Book flashcards` fallbacks. The rule is recorded in the frontend conventions. |
| `336ec28a` | **F6** | Four unlink wordings become **Remove link**, naming the card the button sits on. Saying the same thing in both note lists turned them into a jscpd clone, so they now share a `LinkedNoteList`. |
| `8180f38f` | **F7** | The picker rendered `secondary={note.kind}` — raw API values in lower case. All three sites now go through one `NoteKindChip`, which resolves the kind with `noteKindOf`, so an unknown kind reads "Other" rather than rendering an empty label as the cards did. |
| `a9d5719b` | **F8** | Three phrasings (four on flashcards) become one `FilteredEmptyState`: one sentence and a **Clear filters** button. `useBookTabFilters` gained a `clearFilters` that unsets the shared search and tag params plus whatever else the page names — label and dates on highlights, kinds and chapter on notes. Until now there was no control anywhere that undid a filtered-to-nothing list. |
| `5fa7e2f5` | **F10** | "Questions to think while reading" gains its "about"; "Logout" becomes "Log out"; the library's empty state stops telling a reader whose *search* found nothing to upload highlights, and distinguishes "no books match your search" from "no books yet". "Run text embedding for the library" becomes the button **Index library** under the heading **Search by meaning** — and the rest of that section says "index" too, so the button and the snackbar it raises no longer disagree. |
| `525e100b` | **F11** | Chapterless cards were grouped under "Book flashcards", styled exactly like a real chapter heading. Now **Not in a chapter**. `ChapterGroupedList` hands its `ariaLabel` the chapter rather than only its name, so the group can call its list "Flashcards not in a chapter" instead of "Flashcards in Not in a chapter". |

### Where I read a finding differently

- **F2** asks for "Type throughout". The filter section is plural ("Types") because it sits beside "Chapters", "Tags" and "Bookmarks", which all name the set they filter on, and the drawer tab it duplicates already read "Types". The editor's select picks one value, so it is singular.
- **F5** deliberately ships without a lint rule: a checker cannot tell a proper noun from Title Case, and every book title, author and chapter name in this app is a proper noun. It would fire on data, not on copy.
- **F6** and **F7** each pulled a small shared component out (`LinkedNoteList`, `NoteKindChip`) rather than repeating the corrected copy in two or three places. F6's was forced: making the strings agree created a clone.
- **F5** also fixed two strings from earlier on this branch — "Delete Tag Group" (A6) and "New Note" — which had been left in Title Case to match their siblings while F5 was untriaged.
- **F8** drops the notes page's quoted-query variant ("No notes match “x”.") in favour of the one sentence. The query is still in the search field directly above it.
- **F10** reaches past its own four strings in one place. Renaming the button to "Index library" while the section's snackbars still said "Text embedding complete." would have reproduced the exact defect the finding names, so the whole section uses "index" — the word its own description already used. Hooks, cache events and endpoints keep `embedding`.
- **F11** takes the finding's first option (rename) rather than its second (distinguish visually): a heading that says what it is needs no second signal.

F9 has no entry in the audit note — it is referenced only from the "ten that matter most" table. Every other agreed finding in section F is here.

## Section G

| Commit | Finding | Change |
|---|---|---|
| `914f88a5` | **G2** | A chapter row counted highlights and flashcards; the dialog it opens counts notes too, so a chapter with six notes and no highlights showed nothing. The note count now sits between the other two and counts exactly what the dialog's Notes tab counts, gists included. It reads the unfiltered notes-for-book query, which the book header already issues, so no new request. |
| `f2a8e1ee` | **G3** | `ChapterNav` printed one number chosen by a `countType` prop, so the same chapter read "12 highlights" on one tab and "3 flashcards" on the next. Both counts on both tabs now, joined by the app's middot; a zero is omitted, which keeps the row quiet and still stable. The flashcards page looks highlight counts up from the book's chapters. |
| `383e0caf` | **G5** | The chapter dialog is deep-linkable via `?chapterId=` and had no way to produce that link. It has the same "Copy link" as highlights and notes, built on the structure route so the URL is not a dead link. The toolbar came out of `AIFeature` to do it — only the generate button is an AI feature, and gating the row took copy-link away with it. |
| `22f323cd` | **G6** | `HighlightLabelsList` returned null below two labels, so a reader whose highlights are all one colour never saw the section and never found the label editor — reachable otherwise only through the colour dot inside a highlight dialog. It shows from one label up. |
| `2c00b12a` | **G7** | A highlight with three linked notes looked identical to a bare one. The count now sits beside the flashcard count on every surface the card appears. **No backend change**: the finding expected one, but `NoteWithLinks.highlight_ids` already comes back with the book's notes, so `useNoteCountsByHighlight` inverts it — reading the query the book header already issues. |
| `ba7d3b8e` | **G8** | The stats strip counts bookmarks, already on `book.bookmarks`. Adding it would have printed "1 bookmarks", which every count in the strip already did at one, so pluralisation moved into one `countLabel` in `utils/counts.ts`, shared with the chapter rows and the chapter sidebar. |

### Where I read a finding differently

- **G7** is the one the audit flagged as having a backend cost. It does not: the link data is already on the notes payload. Nothing was added to the API.
- **G2**, **G7** and **G8** each pulled accessibility along with them. An icon beside a bare number reads as "2" to a screen reader, so each count pair is now a `role="img"` labelled "2 notes" — which is also what makes the counts testable.
- **G8** notes that tags are uncounted too; its fix asks only for bookmarks, so that is all it does.
- **G3** omits a zero rather than printing "0 flashcards". The finding wants the row stable across tabs, and omitting zeros keeps that while staying quiet.

G1 and G4 are marked needs-discussion and are untouched.

## Section H

| Commit | Finding | Change |
|---|---|---|
| `743bacb9` `810ac381` | **H2** | `CommonDialog` put previous/next in the footer at every width and `CommonDialogHorizontalNavigation` added a second pair flanking the content from `sm` up, so a tablet carried four arrow buttons for two actions. The flanking pair is gone; the footer's is the only one, at every width. |

### Where I read a finding differently

- The first pass (`743bacb9`) kept one pair per breakpoint — flanking on a wide screen, footer on a phone. **Revised at review (`810ac381`): the footer pair is the only one, at every width.** The room the flanking arrows took stays, though: the content is inset 56px a side above `sm`, so a paging dialog keeps the measure it had. Measured either way at 1440px, the content is 440px wide before and after.
- Getting there meant one component owning the arrows, so `CommonDialogHorizontalNavigation` is deleted. Left as two components, a dialog passing `navigation` without the wrapper would silently lose its arrows above `sm` — and `CommonDialog`'s own test did exactly that. All three real callers used them together anyway.
- The two pairs also disagreed about being disabled: the flanking arrows took their own `disabled` prop, so on the note dialog they stopped during a delete but not during a save, while the footer's stopped during both. What survives follows the dialog's `isLoading`.

H1 is marked needs-discussion and is untouched. H3–H7 were removed from the audit note before implementation.

## Section I

| Commit | Finding | Change |
|---|---|---|
| `32b8b018` | **I1** | The content column was one of three widths depending on the tab: Highlights and Flashcards subdivided their share into content plus a rail, Notes and Reflection capped at 864px, Structure and Sessions ran to the full window. The shell owns the grid now — fixed nav, fixed measure, and a right rail whose column is reserved on every tab whether or not it is filled. The two tabs with a rail portal into it, the way the left sidebar already worked. `ContentWithSidebar` and `MiddleContentColumn` have no callers left. |
| `c035f78f` | **I5** | `m: aiEnabled ? -3.5 : -1` on the session `ul` and `py: aiEnabled ? 3.5 : 1` on the card meant turning AI off moved the page's outer margins. Constant card padding and a plain list reset; the summary block carries its own spacing, which it already did. |

### Where I read a finding differently

- **I1** also makes the content column a `main` landmark, at both widths. The app had none. It is what lets a test say the rail sits beside the content rather than inside it — my first attempt measured `main` alone and passed against a deliberately broken grid, because `main` was the same width on both tabs before this change too.
- **I5** drops the negative margin rather than restating it as a constant. There is no card surface here — no background, no border — so it was only pulling text a few pixels outside the content column, and by a different amount than the padding it was nominally cancelling.

## Section J

| Commit | Finding | Change |
|---|---|---|
| `11a87451` | **J1** | The tag group's title was a clickable `Box` with no role, tab stop or key handler, and its chevron was a bare icon. It is a `ButtonBase` with `aria-expanded` and `aria-controls` now. Chapters and bookmarks, the finding's other two, were already settled by B7. |
| `50cce940` | **J2** | `NoteCard` was a styled `Box` with `role="button"`, `tabIndex` and a hand-written Enter/Space handler that had to check the event target. It uses the shared `HoverableCardActionArea`. That forced the unlink action out of the card body — a button cannot contain one — onto the top corner, which is the pattern the note dialog already uses. `ChapterAccordion`, the finding's other half, became `ButtonBase` under A8/B7. |
| `1e759442` | **J3** | `ReadingSessionCard` put `HighlightCard`s straight in a `ul`, and the outer list's `li` lived *inside* the card. Both go through `CardList`, each list owning its own `li`. Also drops the card's `component` prop, which nothing passed and it never read. |
| `bdb9bdea` | **J4** | The chapter generation attached no error handler in either place it is triggered, so a failure just stopped the spinner. Both attach one, named for what the reader asked for. The finding's third case, `useUpdateDigestAnswers`, already got its handler under A5. |
| `0013700e` | **J5** | Both Settings forms held their own `success` state and rendered an inline `Alert`. Both use the snackbar. It also fixes the password form's worst moment: it set success then called `logout()`, so the alert painted as the page unmounted — the snackbar outlives the route change and now says "Signing you out." |
| `a08d3528` | **J8** | "Press Enter to add a tag, click X to remove" is gone, along with the `helperText` prop that existed only to carry it. |
| `f30a90b2` | **J9** | The chat and quiz shared a hard-coded "Type your answer...". The placeholder is on the variant beside the title it already owns. |

### Where I read a finding differently

- **J5** moves only the success path. The failure alerts stay inline: "Check your current password" belongs beside the field it is about, not in a message that slides away. (The finding's warn-before-signing-out clause was removed from the audit note before implementation.)
- **J2** keeps the note title as an `h3` inside the button, so the notes list keeps its heading structure and the tests that navigate by heading still hold.

I2, I3, I4, G1, G4, H1, D2, D6, E3, A2, A3 and A7 are marked needs-discussion and are untouched. I6, I7, J6 and J7 were removed from the audit note before implementation.

## Tests

Seven new route-level tests. From A–E: tag-group delete confirmation, the reflection answer opening in edit mode, a digest answer reporting its save and not shifting the layout, the chapters section collapsing from the keyboard. From F: the note picker naming a note's type (it fails against the old markup), and Clear filters restoring both notes and an empty search field — with the highlights page's composed four-filter test now clearing all four with the one button. From G: a chapter row's note count, both counts in the chapter sidebar, the chapter dialog's copy-link surviving AI being off, the labels section appearing with one label, a highlight card's note count scoped to the card that owns it, and the stats strip's bookmark count. Four of them fail against the code they replace. From H: one pair of arrows at each width, which fails against the old always-on footer pair. From I and J: the chapter rail living outside `main`, a tag group collapsing from the keyboard, a note card's unlink control not nested inside the card's own button, a failed generation raising a snackbar, and an email update reporting through one. Four of those fail against the code they replace.

145 tests, `type-check`, `lint`, `format:check`, `knip` and `duplication-check` (1.04%, threshold 1.1) all pass.

One test was removed rather than added (`0db9627f`): the flashcard action-clearance check from A5 resized the viewport and measured immediately, which raced a remount across the desktop breakpoint and failed on CI with a zero-width rect.

https://claude.ai/code/session_01K5XULybBdyELdbshcxrGh1


frontend/src/pages/BookPage/common/highlightDates.ts
```diff
@@ -47,13 +48,13 @@ export const filterChaptersByHighlightDate = (
 export const getLastSevenDaysFrom = (today: DateTime<boolean> = DateTime.local()): string =>
   today.minus({ days: 6 }).toFormat(DATE_SEARCH_FORMAT);
 
+/**
+ * The `en-US` pin in `parseStrictly` is for reading the backend's fixed format;
+ * rendering goes through the app's one formatter, in the browser's locale.
+ */
 export const formatHighlightDate = (value: string): string => {
```
Tumetsu, line 64: When does backend return US dates? Backend should only send ISO timestamps.

frontend/src/components/cards/HoverableCardActionArea.tsx
```diff
@@ -1,5 +1,14 @@
...
+ * A list card that opens something.
+ *
+ * The left rail is a type marker, never hover state — a permanent amber rail
+ * means "note" or "flashcard", and a card without one is a highlight. Hover is
+ * said the same way on all three: a background tint and a shadow lift. The
+ * transparent rail here only keeps the text aligned with the cards that have
+ * one.
+ */
 export const HoverableCardActionArea = styled(CardActionArea)(({ theme }) => ({
   borderLeft: `3px solid transparent`,
```
Tumetsu: Border left should be 0 when the color is removed here.

frontend/src/components/cards/HoverableCardActionArea.tsx
```diff
@@ -1,5 +1,14 @@
 import { CardActionArea, styled } from '@mui/material';
 
+/**
```
Tumetsu, line 3: I think this explanation is weird here because the this specific component is not used in contexts described here such as flash cards and notes.

frontend/src/pages/BookPage/common/highlightDates.ts
Tumetsu: You're right that the claim was wrong — the backend never sends a US-formatted date. Corrected in `cb6f7e48` (now `675cd831`).

Two separate things were tangled in that comment:

- **`highlight.datetime` is not ISO.** It is KOReader's own `yyyy-MM-dd HH:mm:ss` string, passed through unnormalised (`highlight_schemas.py:30`, "KOReader datetime format"). Every other timestamp — `created_at`, `start_time` — is ISO and goes through `formatDate`. The comment now says that explicitly rather than "the backend's fixed format".
- **The `en-US` pin is about digits, not date order.** The format is numeric, so nothing about it is locale-dependent to read — but luxon parses and re-renders in the ambient locale's numbering system, so the strict round-trip check underneath breaks. Measured:

| locale | unpinned parse | unpinned round-trip |
|---|---|---|
| en-US, fi-FI | valid | `2026-07-05 23:00:00` |
| ar-EG | valid | `٢٠٢٦-٠٧-٠٥ ٢٣:٠٠:٠٠` → rejected |
| hi-IN-u-nu-deva | **invalid** | — |

So a reader on either locale would have seen every highlight date fall through to the raw string. The pin stays; the comment now gives that as the reason.

frontend/src/components/cards/HoverableCardActionArea.tsx
Tumetsu: Done — the `borderLeft` is gone entirely rather than set to 0, since with no colour to paint it was only reserving 3px of dead space. Folded into `0367be6b` (now `5f321010`).

frontend/src/components/cards/HoverableCardActionArea.tsx
Tumetsu: Agreed — I described the rule across all three card types in the one component that is not any of them. Its three consumers are `HighlightCard`, `AISummary` and `RelatedContentCard`; no note or flashcard card goes through it.

Rewritten to name those contexts and say what this component does, keeping only the part that constrains it — that hover here is a tint and a lift, never a coloured rail. Same commit as the border removal.

frontend/src/pages/BookPage/navigation/HighlightLabelsList.tsx
```diff
@@ -85,7 +85,11 @@ export const HighlightLabelsList = ({
   const { data } = useGetBookHighlightLabels(bookId);
   const labels = data?.items;
 
-  if (!labels || labels.length < 2) {
+  // Shown from one label up. Hiding the section below two meant a reader whose
```
Tumetsu, line 88: remove this comment block

PR #667, opened 2026-08-29
MUI Guidelines based UI/UX improvements

Tumetsu:
Six findings from the MUI-guidelines pass, plus what came out of trying them. Each is its own commit.

## Icon buttons were all small ones

`IconButtonWithTooltip` hard-coded `size="small"` with no way to override it, and `createAdaptiveTouchTarget()` promised "minimum 44px touch target on mobile" in its docstring while setting `minHeight/minWidth: 24`.

The theme now carries the rule once: `MuiIconButton.styleOverrides.root` gives every icon button a 48dp minimum on a coarse pointer — the 44 raw `IconButton size="small"` call sites included, which a fix at the wrapper alone would have missed. It is reset inside `.MuiInputBase-root`, because an Autocomplete's clear and dropdown marks and the date picker's calendar button are `IconButton`s living in a `TextField`, and a 48px minimum on those stretches the field.

`IconButtonWithTooltip` takes a `size` and defaults to MUI's own `medium`. `createAdaptiveTouchTarget` is deleted rather than corrected: the theme covers what it claimed to do, and it had one caller. That caller — the tag-group actions — also had its 2px gap widened to 4px next to the delete.

Two corrections to the finding as written:

- The targets were 30–34px, not 28. MUI's `size="small"` sets `fontSize: 18px` on the button, but `SvgIcon` defaults to `fontSize="medium"` (24px) and does not inherit, so 18px glyphs only appear where a call site passes `fontSize="inherit"` — none of these do. Now 40px and 36px on a mouse, 48px on touch.
- The tag-group header was already partly fixed by the merged audit batch: those icons are `ICON_SIZE.ui` in `text.secondary`, and the delete has opened a `ConfirmationDialog` since `TagGroupSection.tsx:99`. What remained was the 24px target.

The manage-book mark in the book title keeps `size="small"` — it sits inline in the `h1`, where medium padding grows the line box without buying anything.

## The confirmation dialog asked the question MUI names as the anti-pattern

Two of the five confirmations said "Are you sure you want to delete this…", under a title restating what the message already said; the other three each phrased it differently.

`ConfirmationDialog` drops `title` and takes the message alone, as the question the two buttons answer. Every call site is now one shape — the question, plus a second sentence only where something non-obvious follows from a yes:

| | before | after |
|---|---|---|
| flashcard | Delete flashcard / Are you sure you want to delete this flashcard? | Delete this flashcard? |
| highlight | Delete highlight / Are you sure you want to delete this highlight? | Delete this highlight? |
| note | Delete note / Delete note "X"? This cannot be undone. | Delete the note "X"? |
| book | Delete book / Are you sure you want to delete "X"? This will permanently delete the book and all its highlights. | Delete the book "X"? Its highlights go too. |
| tag group | Delete tag group / Delete the group "X"? Its 2 tags stay… | Delete the group "X"? Its 2 tags stay… |

The dialog is also `role="alertdialog"`. MUI v9 has that as a first-class `Dialog` prop, so `slotProps.paper` is not needed. `aria-labelledby` points at the message, since dropping the title leaves nothing else to name the dialog.

## The snackbar covered the mobile bottom navigation

`SnackbarContext` anchored to the bottom edge with no offset, so on a phone every message landed on the book page's bottom navigation for six seconds — the error messages included, which is when a reader is most likely to want to navigate away.

The bottom navigation is the book page's alone, and only below `lg`, so it publishes the room it takes as a custom property while it is mounted, and the snackbar reads that as a `margin-bottom`. Where the property is unset — the library, settings and auth pages, and the book page on `lg` — the margin is `0px` and MUI's own 8px/24px place at the bottom edge stands. Using a margin rather than `bottom` is what keeps those defaults: no MUI breakpoint values copied out to serve as a fallback. The clearance itself moves to `Layouts.tsx` beside `PAGE_GUTTER`, since two pieces of chrome now depend on it.

The property is set from a `useLayoutEffect` because the snackbar's provider sits above the router in both `main.tsx` and the test harness, so no React context can carry the fact upward.

One thing the finding did not cover: the FAB column shares that corner, and the snackbar spans the width below `lg`. Offsetting the snackbar alone would have traded the collision with the navigation for a collision with the add-FAB. So the column rides above an open message and settles back when it closes — which is what Material shows for a snackbar over a FAB. On `lg` the snackbar is centred and narrow and never reaches that far right, so nothing moves there.

## Icon buttons said one thing and answered to another

`IconButtonWithTooltip` took a tooltip `title` and an `ariaLabel`, and MUI spreads the child's own props after the tooltip's — so the `ariaLabel` became the accessible name while the `title` was what the reader saw. Ten of the seventeen call sites had the two differ. Seven were benign, because the visible text was contained in the name ("Copy link" inside "Copy link to highlight"), which is what WCAG 2.5.3 asks for. Three were not:

| visible tooltip | accessible name | where |
|---|---|---|
| Copy highlight content | Copy highlight text | `HighlightViewDialog/components/Toolbar.tsx` |
| Cancel generation | Cancel batch generation | `Structure/BatchDigestToolbar.tsx` |
| View note | View linked note | `Flashcards/FlashcardListCard.tsx` |

The two props collapse into one `label`, so the divergence cannot come back rather than being corrected three times. Where the strings differed the more specific one wins and becomes the tooltip too — "Edit" and "Delete" in a flashcard card are now "Edit flashcard" and "Delete flashcard", which is the better tooltip anyway, and keeps the accessible names the existing tests query. `UnlinkButton` renames its own `title` to match.

I did not add the `describeChild` escape hatch the finding suggested. No call site wants a description that differs from the name, and an unused optional prop on a shared component is the dead API surface `knip` exists to catch. It is three lines when a case turns up.

## The theme now carries what the call sites repeated

Three of MUI's customization mechanisms were going unused while the same intent was restated at each call site.

**`defaultProps`.** All twelve chips in the app passed `size="small"`. The theme says it once and the props go.

**`variants` in `styleOverrides`.** `filterChipSx(isSelected)` held the geometry, transition and hover lift of a chip that selects something, and four chips had to remember to spread it. The theme matches on `ownerState.clickable` instead, so a clickable chip gets the treatment whether or not anyone remembered, and `filterChipStyles.ts` is deleted. The label chip still overrides the colour on top — a highlight label's colour is the one that means something.

**Custom typography variants.** `PageTitle` rendered `variant="h2"` and then corrected the weight and colour in `sx`, which is how the two drifted apart: `h2` is the light subtitle weight a book's author line uses, and a page title is the heavy one. `pageTitle` and `sectionTitle` are theme variants now, with `variantMapping`, and the components keep what a theme cannot hold — the heading level, the spacing below, the gradient rule.

I skipped `MuiIconButton: { defaultProps: { size: 'medium' } }`. `medium` is already MUI's default, so the entry is a no-op; K1 was `IconButtonWithTooltip` hard-coding `small`, and removing that hard-code is what fixed it.

Verified in the browser rather than by type-check, because `variants` matching on an `ownerState` callback either fires or silently does nothing: a clickable chip gets `padding: 2px 4px` and the 0.2s transition, a display chip keeps MUI's own `padding: 0` at height 24, and `pageTitle` renders an `h2` at weight 900 in `#43311E`. That check was a scratch file, run and deleted — it asserts computed CSS, which is not the suite's tier.

## One focus ring, in the theme

Three components hand-rolled the same `:focus-visible` block, and jscpd flagged two of them as a clone. `MuiButtonBase` carries it now, so every clickable thing has one — buttons, icon buttons, chips, menu items, sidebar rows — rather than the three that remembered.

It is drawn **inside** the control's edge (`outlineOffset: -2`). An outset ring reads better on a standalone button, and that is what I built first, but it broke on the two shapes this app is mostly made of:

- a **menu item** is full-bleed inside a rounded `Paper` — its square ring escaped the panel's corner at the top, and a menu with a `maxHeight` would clip it outright;
- a **sidebar row** is a `fullWidth` button in a narrow column, with no room outside it for a ring to live.

Inset never escapes and never clips. The cost is that the ring vanishes on a filled amber surface, so `MuiButton`'s `contained` and the selected-chip variant switch it to `primary.contrastText`.

Verified in Chromium on every control type: [screenshots and the reasoning](https://claude.ai/code/artifact/c1d8d6ef-53eb-455c-899b-ae0ff6b2cbd6).

## Two follow-ups from trying the ring

**The nav swallowed keyboard focus.** Each desktop nav item was a TanStack `Link` wrapping a `ListItemButton` — an anchor with a `role="button"` element inside it. Tab stopped on both, so a keyboard reader crossed every destination twice, and the two stops sit on the same row: the browser's blue ring on the anchor, then the theme's brown one on the button. It reads as the focus being stuck on one item. The row is the link now, via `createLink(ListItemButton)` — the same pattern, and the same reason, as `GlobalSearchResultRow`: `component={Link}` would have `ListItemButton` copy `to` into an `href` that the router then re-parses, dropping search params.

**The section titles lost their font.** A regression from the typography-variant commit above. MUI puts `fontFamily` on every variant it builds itself (`createTypography.js:49`) and merges custom ones straight through, so `sectionTitle` inherited the font — and inside an `AccordionSummary`, which is a `<button>`, that meant the browser's own sans-serif. The collapsible section headers went bold Arial. The stack is a named constant now, stated by `typography.fontFamily` and by both custom variants.

**And the focus ripple is off.** MUI starts a permanently pulsating ripple on keyboard focus (`ButtonBase.js:202`) that `Button` and `IconButton` both opt into — Material Design 1's focus affordance, from before `:focus-visible` existed. With the ring it was a second indicator for the same state. `disableFocusRipple` on both; the click ripple is untouched.

## Review round

Every comment this branch added is gone — the eight flagged inline, then the same pass over the rest. Where a comment already existed on `main` and this branch only edited it, it went back to what `main` said.

Two magic numbers survived that pass and shouldn't have, so they are written out instead of commented:

```ts
/** The bottom navigation's own height, plus air, plus the device's safe area. */
export const BOTTOM_NAV_CLEARANCE = 'calc(56px + 24px + env(safe-area-inset-bottom))';

/** MUI's snackbar offset below `lg`, plus a one-line `Alert`, plus air. */
export const SNACKBAR_CLEARANCE = 'calc(24px + 48px + 8px)';
```

Both were `80px` before, meaning two unrelated sums — the part that would have gone wrong first, since changing the nav's height gives no reason to look at the snackbar constant. `BookPage.tsx` composes them into a nested `calc()`, checked in Chromium: 80px, 80px, 160px composed, the same pixels as before.

`98b4bce3` (dialog horizontal padding, 7 → 2) is yours, on the branch.

## Verification

`type-check`, `lint`, `knip`, `duplication-check` (1.0%, threshold 1.1) and `build` all pass, and 145 tests. Two `TagsList` confirmation tests moved from `getByRole('dialog')` to `getByRole('alertdialog')` — the role change showing up where it should, and the only test edit on the branch.

Three things were verified in a real browser rather than by type-check, because each either fires or silently does nothing, and then deleted — they assert computed CSS, which is not the suite's tier:

| what | result |
|---|---|
| the chip `variants` callback and the custom typography variants | clickable chip `padding: 2px 4px` + 0.2s transition; display chip keeps MUI's `padding: 0` at height 24; `pageTitle` renders `h2` at weight 900 in `#43311E` |
| the focus ring | [screenshots of every control type](https://claude.ai/code/artifact/c1d8d6ef-53eb-455c-899b-ae0ff6b2cbd6) |
| the nav's tab stops and the composed clearances | one anchor per destination with nothing focusable inside it; 80px / 80px / 160px |

https://claude.ai/code/session_01Fupr32CuL6MNRw2rubiqyU


frontend/src/components/buttons/IconButtonWithTooltip.tsx
```diff
@@ -2,10 +2,15 @@ import { IconButton, Tooltip, type IconButtonProps } from '@mui/material';
 import type { ReactNode } from 'react';
 
 interface IconButtonWithTooltipProps {
-  title: string;
+  /**
```
Tumetsu: remove this comment block

frontend/src/theme/theme.ts
```diff
@@ -273,6 +307,61 @@ export const theme = createTheme({
         },
       },
     },
+    MuiTypography: {
+      defaultProps: {
+        // Custom variants have no element of their own without this, so a page
```
Tumetsu: remove this comment

frontend/src/theme/theme.ts
```diff
@@ -273,6 +307,61 @@ export const theme = createTheme({
...
+        // title written as `<Typography variant="pageTitle">` would render a
+        // paragraph.
+        variantMapping: {
+          pageTitle: 'h2',
+          sectionTitle: 'h2',
+        },
+      },
+    },
+    MuiChip: {
+      // Every chip in the app is a small one, from a tag to a filter to the
```
Tumetsu: remove comment

frontend/src/theme/theme.ts
```diff
@@ -273,6 +307,61 @@ export const theme = createTheme({
...
+      // "not on device" marker.
+      defaultProps: { size: 'small' },
+      styleOverrides: {
+        root: ({ theme: t }) => ({
+          variants: [
+            {
+              // A chip you can click selects something — a tag, a label, a
```
Tumetsu: remove comment

frontend/src/components/typography/PageTitle.tsx
```diff
@@ -10,14 +10,13 @@ interface PageTitleProps {
   component?: 'h1' | 'h2';
 }
 
-/** The heading every page leads with. Inherits `text-align` from its parent. */
+/**
```
Tumetsu: remove comment

frontend/src/theme/theme.ts
```diff
@@ -20,6 +20,14 @@ const colors = {
  * Custom colors used throughout the application.
  * These are consolidated from various rgba() calls in components.
  */
+/**
```
Tumetsu: remove comment

frontend/src/theme/theme.ts
```diff
@@ -295,6 +305,10 @@ export const theme = createTheme({
       },
     },
     MuiButton: {
+      // The ring above is the focus indicator; MUI's pulsating focus ripple is
```
Tumetsu: remove comment

frontend/src/pages/BookPage/navigation/DesktopNavLinks.tsx
```diff
@@ -1,7 +1,19 @@
 import { Box, List, ListItemButton, ListItemIcon, ListItemText } from '@mui/material';
-import { Link, useMatchRoute } from '@tanstack/react-router';
+import { createLink, useMatchRoute } from '@tanstack/react-router';
 import { BOOK_PAGE_LABELS, BOOK_PAGE_ROUTES } from './bookPageRoutes.ts';
 
+/**
```
Tumetsu: remove comment

frontend/src/components/buttons/IconButtonWithTooltip.tsx
Tumetsu: Removed in 3330611e.

frontend/src/theme/theme.ts
Tumetsu: Removed in 3330611e.

frontend/src/theme/theme.ts
Tumetsu: Removed in 3330611e.

frontend/src/theme/theme.ts
Tumetsu: Removed in 3330611e.

frontend/src/components/typography/PageTitle.tsx
Tumetsu: Removed in 3330611e.

frontend/src/theme/theme.ts
Tumetsu: Removed in 3330611e.

frontend/src/theme/theme.ts
Tumetsu: Removed in 3330611e.

frontend/src/pages/BookPage/navigation/DesktopNavLinks.tsx
Tumetsu: Removed in 3330611e.

PR #682, opened 2026-08-29
Show highlight, note and flashcard counts on library book cards

### LINKED ISSUE #672: Show highlight and flashcard counts on library book cards
From the Crossbill UI audit — finding **G1**, severity `high`. Part of #661.

`LandingPage/components/BookCard.tsx`, `api/generated/model/bookWithHighlightCount.ts`

## Problem

`BookWithHighlightCount` carries `highlight_count` and `flashcard_count` on every card, and neither is rendered anywhere. The only use of `highlight_count` in `src/` is on highlight *labels*. `BookCard`'s own comment still describes "reading-stage and highlight-count overlays"; only the stage overlay survives.

On the main screen of the app, a book with 400 highlights and one with none are visually identical apart from a small stage marker.

## Fix

A single muted line under the author: `412 highlights · 38 flashcards`. No new endpoint needed — the data is already on the wire.

## Discussion

Placement and wording on the card are open; the audit did not settle whether counts belong under the author or as an overlay.

Tumetsu:
Closes #672. Design was settled in [a session on the ticket](https://github.com/Crossbill-App/crossbill-web/issues/672#issuecomment-5463289493).

## What changed

Each library card carries a muted row under the author — highlights, notes, flashcards, as icons and numbers, in that fixed order. A count of zero is dropped; a book nobody has marked up shows no row at all. Both surfaces get it, since `BookCard` is shared by the all-books grid and the Recent books carousel.

Words don't fit a 150px card (`412 highlights · 38 flashcards` is ~195px), and the icon vocabulary already exists in `theme/Icons.tsx` and on the book-detail chapter rows, so a reader who has seen a chapter row reads the card for free.

Counts are not clickable: the card is already a `<Link>`, and nested anchors would mean restructuring it for one saved tap.

## Notes are new on the wire

The ticket assumed all three numbers were already available; only two were. `note_count` is a new correlated subquery through `note_books`, scoped to the book's owner and excluding gists — matching `DEFAULT_NOTE_KINDS`, so the card and the Notes tab agree. The exclusion lives as `UNCOUNTED_NOTE_KINDS` beside `NoteKind` in the domain rather than as a string literal in SQL.

`notes.kind` is nullable and untyped notes read as `other`, so the filter is a null check plus `NOT IN`, not a bare `!=` — the latter silently drops every untyped note.

## A pre-existing undercount, fixed

`BookStatsStrip` summed `book.chapters[].highlights.length`, but `book_details_query` drops highlights with `chapter_id IS NULL`, which the upload path leaves unset when the plugin sends no `chapter_number`. So the book page already undercounted, and putting a correct count on the card would have made the mismatch visible. `BookDetails` now carries a real `highlight_count`.

Surfacing those highlights in the chapter tree is a separate issue — they are currently unreachable from that view.

## Duplication

`CountWithIcon` had already been written twice: once in `ChapterAccordion`, once as `CountBadge` in `HighlightCard`. It moves to `src/components/` and both call sites migrate to it. It renders inline now, so it works inside a `MetadataRow` as well as in a flex row.

## Verification

930 backend tests, 147 frontend tests, pyright clean, all five import-linter contracts kept, knip and lint clean. Duplication unchanged (1.02% frontend, 2.30% backend), so no threshold moves.

Each new test was mutation-checked — code broken, failure confirmed, reverted. The user-isolation test earned its keep: it caught a missing owner filter on the note subquery, where another user's note linked to your book inflated the count.

No `koreader-plugin` minimum bump: both new fields are additive.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -4635,6 +4635,13 @@
             "title": "Chapters",
             "description": "List of chapters with highlights"
           },
+          "highlight_count": {
+            "type": "integer",
+            "minimum": 0.0,
+            "title": "Highlight Count",
+            "description": "Live highlights on the book, including any that sit in no chapter",
+            "default": 0
+          },
           "reading_position": {
             "anyOf": [
               {
@@ -5010,6 +5017,13 @@
             "description": "Number of flashcards for this book",
             "default": 0
           },
+          "note_count": {
+            "type": "integer",
+            "minimum": 0.0,
+            "title": "Note Count",
+            "description": "Number of notes for this book, gists aside",
+            "default": 0
+          },
           "reading_stage": {
             "anyOf": [
               {
@@ -5088,7 +5102,7 @@
           "updated_at"
         ],
         "title": "BookWithHighlightCount",
-        "description": "Schema for Book with highlight and flashcard counts."
+        "description": "Schema for Book with highlight, flashcard and note counts."
       },
       "Bookmark": {
         "properties": {
--- a/backend/src/application/library/queries/book_details.py
+++ b/backend/src/application/library/queries/book_details.py
@@ -57,6 +57,7 @@ class BookDetailsView:
     bookmarks: tuple[BookmarkView, ...]
     book_flashcards: tuple[FlashcardRef, ...]
     chapters: tuple[ChapterWithHighlightsView, ...]
+    highlight_count: int
     reading_position: Position | None
     end_position: Position | None
     created_at: dt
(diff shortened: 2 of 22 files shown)


Tumetsu:
Design change after seeing it rendered: the counts moved from a muted line under the author onto the cover itself.

As secondary text under the author they read as more of the same and disappeared into the card. They now sit on a dark scrim, edge to edge across the foot of the cover, white on it whatever the artwork underneath. The strip exists only when there is a count to show, so an unmarked book keeps a clean cover rather than an empty band.

A primary-coloured pill was tried first and dropped: it overflowed the 150px cover at four-digit counts, and its amber competed with the reading-stage marker, which is the one accent that should stand out up there. A full-width strip has the whole cover width, so `1204 · 12 · 137` fits with room to spare.

`coverScrim.strip` is a new token in the theme's `customColors`. The existing translucent whites are documented as belonging only over the amber app bar, "where what sits behind them is known and fixed" — a scrim over cover art is the opposite case, and darkening the unknown is the point.

PR #686, opened 2026-08-30
Fix the mobile filter drawer leaving the page unscrollable

Tumetsu:
### Scrolling stops after setting a filter

Setting a filter on mobile left `<body>` stuck at `overflow: hidden` — the page stopped scrolling and nothing short of a reload brought it back.

Two scroll locks were fighting over the body. The drawer holds our reference-counted `useBodyScrollLock`, but MUI's `Modal` runs one of its own that nothing turned off. MUI records the styles it will restore when its *portal* mounts, which is a render **after** our lock has already pinned the body — so it took our `overflow: hidden` for the page's own value and dutifully put it back on close.

Traced with a `MutationObserver` on `<body>`:

```
--- clicking a tag ---
overflow=<empty> position=<empty> top=<empty>   ← our unlock, correct
overflow=hidden  position=<empty> top=<empty>   ← MUI restoring our own lock
```

That shape matches the symptom exactly: `overflow: hidden` without `position: fixed`, so the page never visibly jumped — it just stopped scrolling.

The fix is the prop `CommonDialog` already passes: `disableScrollLock` on the `Drawer`, leaving our lock as the body's single owner. The reference counting itself was never at fault.

This was not specific to highlights — Notes and Flashcards use the same drawer and had the same bug.

### The filter count sat behind the FAB

MUI puts `zIndex: theme.zIndex.fab` (1050) on the `Fab` root but only `zIndex: 1` on the badge, both inside the badge's own stacking context, so the button painted over its own count. Lifted the badge to `theme.zIndex.fab + 1`.

## Tests

- `FilterDrawer.test.tsx` — picking a tag, and closing by hand, both assert the body is fully unlocked. The check waits for the modal to leave the DOM: MUI restores its bookkeeping only when the exit transition ends, so an earlier assertion races the very restore under test (my first version passed for that reason).
- `FilterFab.test.tsx` — `elementFromPoint` at the badge's centre, so the assertion is that the count is genuinely on top rather than that a style string is present.

Both fail without their fix. Full frontend suite passes (151 tests); lint, type-check and jscpd (1.0%, threshold 1.1) clean. Verified by hand in the dev server.

https://claude.ai/code/session_01FFqeXVpUvLgcYmyobDkwvf

### DIFF
--- a/frontend/src/pages/BookPage/common/FilterFab.tsx
+++ b/frontend/src/pages/BookPage/common/FilterFab.tsx
@@ -13,7 +13,12 @@ export const FilterFab = ({ activeFilterCount, onClick }: FilterFabProps) => {
 
   return (
     <Zoom in={true} mountOnEnter unmountOnExit>
-      <Badge badgeContent={activeFilterCount} color="secondary" overlap="circular">
+      <Badge
+        badgeContent={activeFilterCount}
+        color="secondary"
+        overlap="circular"
+        slotProps={{ badge: { sx: { zIndex: (theme) => theme.zIndex.fab + 1 } } }}
+      >
         <Fab
           size="small"
           color={isFiltered ? 'primary' : 'default'}
--- a/frontend/src/pages/BookPage/navigation/FilterDrawer.tsx
+++ b/frontend/src/pages/BookPage/navigation/FilterDrawer.tsx
@@ -26,7 +26,7 @@ export const FilterDrawer = ({ open, onClose, tabs, header }: FilterDrawerProps)
   };
 
   return (
-    <Drawer anchor="bottom" open={open} onClose={handleClose}>
+    <Drawer anchor="bottom" open={open} onClose={handleClose} disableScrollLock>
       {/* Capped against the *dynamic* viewport, so the drawer stays clear of
           mobile browser chrome however much of it is showing. It scrolls its
           own content; `contain` keeps a pull that runs past either end from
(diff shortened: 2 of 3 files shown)


frontend/src/pages/BookPage/navigation/FilterDrawer.tsx
```diff
@@ -26,7 +26,11 @@ export const FilterDrawer = ({ open, onClose, tabs, header }: FilterDrawerProps)
   };
 
   return (
-    <Drawer anchor="bottom" open={open} onClose={handleClose}>
+    // `disableScrollLock` because `useBodyScrollLock` above owns the body:
```
Tumetsu: remove this comment

frontend/src/pages/BookPage/common/FilterFab.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { FilterFab } from '@/pages/BookPage/common/FilterFab';
```
Tumetsu: no need for this test

frontend/src/pages/BookPage/common/FilterFab.tsx
```diff
@@ -13,7 +13,15 @@ export const FilterFab = ({ activeFilterCount, onClick }: FilterFabProps) => {
 
   return (
     <Zoom in={true} mountOnEnter unmountOnExit>
-      <Badge badgeContent={activeFilterCount} color="secondary" overlap="circular">
+      {/* MUI gives the Fab `zIndex: fab` but the badge only `zIndex: 1`, so
```
Tumetsu: remove this comment

PR #706, opened 2026-08-31
Add a reading-activity grid to the book statistics tab

### LINKED ISSUE #704: Add activity chart to the statistics page
Calculate the book activity chart data on the backend, serve it via the statistics API and render it with [react-activity-calendar](https://github.com/grubersjoe/react-activity-calendar).

A **Reading activity** section on the book's statistics tab, between the reading summary and the session list: a year of squares, one per day, darker the more of the book that day got through.

## What a day counts

- **Pages**, when every session of the book recorded both `start_page` and `end_page`; **minutes** otherwise. All-or-nothing on purpose — under any softer rule a book synced partly from a device that reported pages would render its page-less days blank and undercount the reading.
- A session counts wholly towards the calendar day it **began** on, in the reader's own timezone, so reading past midnight lands on the day the reader sat down.
- Level 0 means `value == 0`, whatever the cause. A day the book was open but no page turned looks the same as a day it was not opened.

## How the four shades divide

Per book, never against global page counts, so a dense book nobody gets through quickly still shows its good days as good days.

Let `t` be the **median** of the window's active-day values:

| Day value | Level |
|---|---|
| `v ≤ 0.5t` | 1 (palest) |
| `v ≤ t` | 2 |
| `v ≤ 1.5t` | 3 |
| `v > 1.5t` | 4 (darkest) |

The median rather than the mean: one all-day binge would otherwise drag `t` up and flatten a year of ordinary reading into a single shade. A book read on exactly one day comes out mid-scale rather than at either extreme.

## Which days the grid covers

A rolling **365-day window** ending on an anchor:

- the **reader's today**, when the book has any session in the last 365 days — so an unread fortnight shows as the gap it is;
- otherwise the **book's last reading day** — so a book finished two years ago still shows when it was read, instead of a year of nothing.

A session dated in the future falls outside the range ending today and so anchors the window on itself, which keeps a skewed device clock from pushing the reading off the grid.

The window is shown as a caption above the grid, with its unit: `Sep 2025 – Aug 2026 · pages read`.

## API

`GET /api/v1/books/{book_id}/statistics?tz=` gains an `activity` object. Sparse — the days with something to show plus the two bounds, roughly 1 KB rather than 11, because the summary above the grid is what the reader sees first and should not wait on 365 mostly-empty entries. The client adds the bounds back and the calendar fills the gaps itself.

```json
"activity": {
  "unit": "pages",
  "range_start": "2025-09-01",
  "range_end": "2026-08-31",
  "days": [{ "date": "2026-03-02", "value": 40, "level": 3 }]
}
```

`activity` is `null` for a book with no sessions, or one whose every day in the window nets zero; the section then keeps its heading and says "No reading activity recorded yet."

## Rendering

- Colours interpolate between two theme tokens — stone-100 for an unread day, `primary.main` for the heaviest — so the ramp follows the brand rather than restating it.
- `colorScheme` is pinned to light; left alone the calendar reads the reader's OS setting and swaps its own grey dark ramp into a light-only app.
- Month names and the first day of the week come from the reader's locale, like every other date the app renders.
- Colour legend on, per-day hover tooltips on, no total-count line.
- One fixed block size, scrolling sideways in a column too narrow for it. Shrinking the squares to fit does not work: a month label is as wide as its name, not as wide as the columns beneath it, so at phone widths the months collide into an unreadable row — and the squares fall under the app's 48px touch minimum long before a year fits.
- The weekday rail is off. The calendar draws those labels at negative x inside the SVG and pushes the SVG clear with a left margin, so they ride along with the grid rather than being pinned beside it: scroll right and they slide under the container's edge half a letter at a time.

## Changed from the original plan

Three rules in the first draft of this ticket were replaced while building it:

1. **A rolling 12-month window replaces the calendar-year grid** and the three year-selection rules with it. A calendar year leaves four months of empty future squares in August and eleven in January; the anchored window ends where the reading does.
2. **The median of active days replaces the mean page count of sessions.** The original mixed units — squares are days, but the average was over sessions, so any multi-session day would have landed in the top bucket.
3. **A minutes fallback exists at all**, for books KOReader syncs by xpoint alone. The original assumed page numbers were always there; they are optional in the sync contract.

Tumetsu:
Closes #704.

A **Reading activity** section on the book's statistics tab, between the reading summary and the session list: a year of squares, one per day, darker the more of the book that day got through.

## The rules

| Rule | Decision |
|---|---|
| Unit | Pages when **every** session of the book recorded both page numbers, minutes otherwise |
| Day value | Sum of the day's sessions, bucketed on the day each **began**, in the reader's timezone |
| Levels | `t` = median of active-day values; `v ≤ 0.5t → 1`, `≤ t → 2`, `≤ 1.5t → 3`, `> 1.5t → 4` |
| Level 0 | `value == 0`, whatever the cause |
| Window | 365 days ending on today, or on the last reading day for a book left longer ago |
| Nothing to plot | `activity: null`; the section keeps its heading and says so |

Two choices worth the reviewer's attention:

**The median, not the mean.** One all-day binge drags the mean up and flattens a year of ordinary reading into a single shade. On `[3, 8, 12, 15, 20, 22, 40, 400]` the mean is 65 and six distinct days all come out level 1; the median is 17.5 and they spread across three shades. A book read on exactly one day also comes out mid-scale rather than darkest.

**All-or-nothing on pages.** Sessions may carry only xpoints — the sync contract requires *either* xpoints *or* pages (`reading_session_schemas.py:56`). Any softer rule would render a mixed book's page-less days blank and silently undercount the reading.

## API

`GET /api/v1/books/{book_id}/statistics?tz=` gains an `activity` object rather than getting a new endpoint. Sparse — the days with something to show plus the two bounds, ~1 KB instead of ~11 — because the summary above the grid is what the reader sees first and should not wait on 365 mostly-empty entries. The client adds the bounds back and the calendar fills the gaps itself.

```json
"activity": {
  "unit": "pages",
  "range_start": "2025-09-01",
  "range_end": "2026-08-31",
  "days": [{ "date": "2026-03-02", "value": 40, "level": 3 }]
}
```

Not a plugin-gated route, so no `koreader-plugin` minimum bump. `backend/openapi.json` is regenerated in the same commit.

## Backend shape

`ReadingActivityCalculator` (domain) owns unit selection, day bucketing, the window anchor and the level rule. `ReadingStatisticsCalculator` takes it by injection and fills a new `activity` field, so the value object keeps one producer and the query keeps one call.

`today` is resolved once in a `reader_today` router dependency and handed down — nothing below the router reads the clock, which is what makes the window a test can assert on.

`ReadingStretch` moved to its own module and gained page numbers. Both calculators are fed the same stretches and must bucket them onto the same days, so leaving it in either service would have made the two import each other.

## Frontend

New dependency `react-activity-calendar` (MIT, React 18||19), which pulls in `date-fns@^4` and `@floating-ui/react@^0.27` — a second date library beside luxon and a second floating stack beside MUI's popper.

Colours interpolate between two theme tokens (stone-100 → `primary.main`) added to `customColors`. `colorScheme` is pinned to light or the calendar reads the OS setting and swaps its grey dark ramp into a light-only app. Months and week start come from the reader's locale.

The grid keeps one block size and scrolls sideways in a narrow column. Shrinking to fit was tried and reverted: a month label is as wide as its name, not as wide as the columns beneath it, so Finnish month names collide into an unreadable row at phone widths.

The weekday rail is off. The calendar draws those labels at negative x inside the SVG and pushes the SVG clear with a left margin, so they ride along with the grid rather than being pinned: scroll right and they slide under the container's edge half a letter at a time. Dropping the rail also brings the grid to 844px, back inside the desktop column, so that width stops scrolling at all.

## Trade-offs accepted

1. On a phone no number is readable without a tap — there is no total-count line, and tooltips need hover.
2. A day read for an hour with no page turns is indistinguishable from a day the book was not opened.
3. Reading more than 12 months before the anchor is invisible; a book read across three years shows only its last year.
4. The statistics response now carries chart data the summary above it does not use, and is time-dependent for actively-read books — any future HTTP caching on this route needs a day-scoped key.

## Verification

- `uv run pytest` 976 passed; `uv run pyright` 0 errors; `uv run lint-imports` 5 contracts kept
- `npm run test` 172 passed; type-check, lint and format clean
- Duplication: backend 2.3%; frontend 1.00% → 0.95%, threshold ratcheted 1.1 → 1.05
- Both new rule sets were mutation-checked: median→mean, anchor→always-today, all→any pages, and dropping the window bounds each fail the suite


frontend/src/pages/BookPage/ReadingStatistics/ReadingActivitySection.tsx
```diff
@@ -0,0 +1,179 @@
...
+export const ReadingActivitySection = ({ bookId }: ReadingActivitySectionProps) => {
+  const { data } = useGetBookStatistics(bookId, { tz: browserTimeZone() });
+
+  if (!data) {
+    return null;
+  }
+
+  return (
+    <Box sx={{ mb: 4 }}>
+      <SectionTitle showDivider>Reading activity</SectionTitle>
```
Tumetsu: We probably can get rid of this section title. This grid is also part of statistics so it belongs to the section on top of it.

frontend/src/pages/BookPage/ReadingStatistics/ReadingStatisticsPage.tsx
```diff
@@ -47,6 +48,8 @@ export const ReadingStatisticsPage = () => {
 
       <ReadingStatsSection bookId={book.id} />
 
+      <ReadingActivitySection bookId={book.id} />
```
Tumetsu: I suppose this could be merged to the ReadingStatsSection.

frontend/src/pages/BookPage/ReadingStatistics/ReadingActivitySection.tsx
```diff
@@ -91,35 +97,55 @@ const RangeCaption = ({ activity }: ActivityGridProps) => {
...
- * minimum long before a year fits. At a fixed size the grid stays legible and
- * tappable, and the overflow is contained here rather than left to nudge the
- * whole page sideways.
+ * minimum long before a year fits.
+ *
+ * The scrolling itself is the calendar's own — it ships a scroll container
+ * around its SVG. Wrapping that in a second one only nests two scrollers and
+ * clips the right edge, so this box sizes and pads rather than scrolls.
+ *
+ * The weekday rail stays off, because it cannot survive that scrolling: the
```
Tumetsu: remove this explanation on weekdays.

frontend/src/pages/BookPage/ReadingStatistics/ReadingActivitySection.tsx
Tumetsu: Do not explain options which were considered. I think this whole comment block can be removed.

frontend/src/pages/BookPage/ReadingStatistics/ReadingStatisticsPage.test.tsx
```diff
@@ -97,36 +98,83 @@ test('the summary stays away until the book has been read', async () => {
...
 });
+
+test('the activity grid draws the whole window, not only the days that were read', async () => {
+  const screen = await renderStatisticsTab({ ...oneSession, statistics: aBookStatistics() });
+
+  await expect.element(screen.getByRole('heading', { name: 'Reading activity' })).toBeVisible();
+
+  // The API sends three days and the window's two ends; every day between them
+  // is the client's to fill, so a 365-day window must draw 365 squares.
+  expect(screen.container.querySelectorAll('[data-date]')).toHaveLength(365);
```
Tumetsu: **Testing standard:** `frontend/claude.md` requires queries by accessible role and name, and recommends adding accessible semantics when they are missing. These assertions instead couple the behavior test to the third-party calendar DOM through `data-date` and `data-level`. Could we give the cells accessible semantics via `renderBlock`, then assert by role/name or exercise the visible tooltip behavior?

frontend/src/pages/BookPage/ReadingStatistics/ReadingActivitySection.tsx
```diff
@@ -0,0 +1,203 @@
...
+ * The reading-activity grid: a year of squares, one per day, darker the more
+ * of the book that day got through.
+ *
+ * Reads the same statistics query as the summary above it, so the grid costs
+ * no second request. While that query is loading or failing there is nothing
+ * to say here — the summary raises the failure — so the section renders only
+ * once an answer has arrived.
+ */
+export const ReadingActivitySection = ({ bookId }: ReadingActivitySectionProps) => {
+  const { data } = useGetBookStatistics(bookId, { tz: browserTimeZone() });
```
Tumetsu: **Non-blocking — duplicated query ownership:** This repeats the same statistics query made by `ReadingStatsSection`, while only that sibling owns error reporting. Would it be cleaner to lift the query into `ReadingStatisticsPage` and pass its state to both sections, so fetching and failure ownership stay in one place?

backend/src/domain/reading/services/reading_activity_calculator.py
```diff
@@ -0,0 +1,174 @@
...
+    MINUTES = "minutes"
+
+
+@dataclass(frozen=True)
+class ActivityDay:
+    """One coloured square: a day, what it counts, and how dark it is."""
+
+    day: date
+    value: int
+    level: int
```
Tumetsu, line 41: **Non-blocking — primitive obsession:** Activity level is a bounded domain concept (1–4 for emitted days), but it is represented as an unrestricted `int` here and in the response schema. Would an `ActivityLevel` enum/value type, plus a constrained Pydantic field, make invalid shades unrepresentable at the domain and API boundaries?

backend/src/application/reading/queries/book_statistics.py
```diff
@@ -17,11 +17,14 @@ class BookStatisticsQueryProtocol(Protocol):
     """Port for reading what a book's reading sessions add up to."""
 
     async def get_statistics(
-        self, book_id: BookId, user_id: UserId, zone: tzinfo
+        self, book_id: BookId, user_id: UserId, today: date, zone: tzinfo
```
Tumetsu, line 20: **Non-blocking — data clump:** `today` and `zone` now travel together through the router, use case, query adapter, statistics calculator, and activity calculator, and `today` is meaningful only relative to that zone. Is this relationship worth capturing in a small reader-calendar context/value object?

Tumetsu (commented):
Standards and spec review complete. The implementation matches issue #704 with no spec gaps. I found one documented test-policy issue and three non-blocking design considerations, noted inline.

backend/src/domain/reading/services/reading_activity_calculator.py
```diff
@@ -0,0 +1,174 @@
...
+        """The last day of the window.
+
+        Today, for a book read within the past year -- so the grid ends on the
+        reader's own today and an unread fortnight shows as the gap it is.
+        For a book last read before that, the window ends on its final reading
+        day instead, rather than showing a year of nothing. A session dated in
+        the future falls outside the range ending today, and so anchors the
+        window on itself.
+        """
+        last_day = max(day_in(stretch.start_time, zone) for stretch in stretches)
```
Tumetsu, line 146: **A future-dated session can crop or empty the grid** (medium)

`max(...)` runs over every stretch, so a session dated in the future falls outside `window_start..today` and becomes the anchor, dragging the window forward with it. The docstring says this is deliberate, so the question is whether it is the behaviour you want:

- Skew of a few weeks: the oldest weeks of real reading fall off the left edge.
- Skew over 365 days: no real reading day is inside `range_start..range_end`, `active` comes out empty, `calculate` returns `None`, and the section renders "No reading activity recorded yet." directly under a summary still reporting session count, hours read and reading span.

Nothing bounds `start_time` on the write path — `ReadingSessionSyncItem` validates only that xpoints or pages are present — so a KOReader device with a reset clock produces exactly this. Clamping the anchor to `today` when `last_day` is more than a window ahead keeps the real reading visible and costs nothing in the legitimate case.

backend/src/domain/reading/services/reading_activity_calculator.py
```diff
@@ -0,0 +1,174 @@
...
+            for day in sorted(active)
+        )
+        return ReadingActivity(
+            unit=unit,
+            range_start=range_start,
+            range_end=range_end,
+            days=days,
+        )
+
+    def _unit(self, stretches: Sequence[ReadingStretch]) -> ActivityUnit:
```
Tumetsu: **The unit is decided on sessions the grid never draws** (low)

`_unit` scans every stretch; `calculate` filters to the window afterwards. A book whose pre-window history came from a device that sent xpoints only, but whose entire visible year recorded pages, is labelled and shaded in minutes — the caption reads "· minutes read" while every session actually on the grid has page numbers.

The all-or-nothing rationale in the docstring — page-less days would render blank — applies only to sessions inside the window, and those are already excluded by the `range_start <= day <= range_end` filter. Filtering to the window before picking the unit would match what gets drawn.

backend/src/domain/reading/services/reading_stretch.py
```diff
@@ -0,0 +1,64 @@
...
+    def pages_read(self) -> int:
+        """Pages the session covered, or zero when it recorded none.
+
+        Never negative: the write side rejects a session that ends on an
+        earlier page than it starts, and a legacy row that slipped through
+        should cost the reader one pale square rather than a broken scale.
+        """
+        if self.start_page is None or self.end_page is None:
+            return 0
+        return max(0, self.end_page - self.start_page)
```
Tumetsu, line 53: **The docstring's premise does not hold, and re-reading days vanish** (low)

"the write side rejects a session that ends on an earlier page than it starts" — it does not. `ReadingSessionSyncItem` (`backend/src/infrastructure/reading/schemas/reading_session_schemas.py:48-61`) checks only `ge=0` on each page and that one of the xpoint/page pairs is present. There is no ordering check anywhere on the write path.

So `start_page=115, end_page=102` from a day spent re-reading earlier chapters is ordinary input, not a legacy row that slipped through — and it contributes 0, so the day is drawn as one the book was never opened. If no day in the window nets a positive delta, the section says "No reading activity recorded yet." beside a summary reporting hours read.

Either use `abs(end_page - start_page)`, or add the ordering validator the docstring assumes exists. As it stands the comment asserts a guarantee the code does not have.

frontend/src/pages/BookPage/ReadingStatistics/ReadingActivitySection.tsx
```diff
@@ -0,0 +1,203 @@
...
+ * through.
+ */
+const useCalendarLocale = () =>
+  useMemo(() => {
+    const locale = browserLocale();
+
+    return {
+      locale,
+      months: Info.months('short', { locale }),
+      weekStart: (Info.getStartOfWeek({ locale }) % 7) as DayIndex,
```
Tumetsu: **Week start does not follow the locale on Firefox** (cosmetic)

Luxon's `Locale.getWeekSettings()` (`node_modules/luxon/src/impl/locale.js:536`) returns the hard-coded `fallbackWeekSettings` (`firstDay: 1`) whenever `Intl.Locale.prototype.getWeekInfo`/`weekInfo` is unavailable — which is Firefox's situation. A US reader on Firefox gets a Monday-start grid.

Not worth changing the code over; the comment above just claims more than the call delivers.

Tumetsu (commented):
Reviewed the backend calculator, the query/router wiring, the frontend section, and the `react-activity-calendar` props contract.

Four findings inline: one medium (a future-dated session can empty the grid), two low, one cosmetic. Nothing that blocks on its own, but finding 3 rests on a guarantee the write side does not actually provide.

Everything else checked out: the `theme.light` pair maps correctly onto the library's `[colorZero, colorPos]` convention; the Luxon 1–7 → library 0–6 `weekStart` conversion is right; `withWindowBounds` always yields sorted, non-empty, in-range data, so `validateActivities`/`fillHoles` cannot throw; every caller of the widened `get_statistics` and `ReadingStatisticsCalculator` signatures was updated, including `protocol_conformance.py` and the DI container; `sessions[-1][2]` still picks the latest `end_position` under the widened `SessionRow`; `openapi.json` matches the schemas; and the jscpd threshold ratcheted down (1.1 → 1.05) as required.

PR #708, opened 2026-09-01
Split library browsing into its own page

Tumetsu:
The landing page now focuses on the reader's dashboard — showing recent books and a link to browse the full library — while all-books search and pagination moved to a dedicated `/library` route.

## Changes

- **New Library page** (`/library`): Hosts the searchable, paginated book list that previously lived on the dashboard. Query parameters (`search`, `page`) stay in the URL for linkability and back-button support.

- **Simplified Landing page**: Now shows only recent books and a "Browse the library" button. Removed search, pagination, and the full book list.

- **Global navigation refactored**:
  - New `AppBarNavLinks` component renders top-level destinations (Home, Library) as an underlined row in the app bar on `md` and up.
  - New `AppBarDrawer` component moves the same destinations into a left drawer on narrow viewports, plus account actions (Settings, Log out) that were previously behind an account icon menu.
  - New `appRoutes.ts` defines the app's top-level destinations in one place, read by both the nav links and drawer.
  - Menu icon added to the app bar to open the drawer on mobile.

- **New layout components**:
  - `SidebarLayout`: A three-column grid (left rail, content, optional right rail) with fixed column widths, extracted from the book page's own layout logic to avoid reflow as tabs change.
  - `BackToLibraryLink`: A back button to the library, shown only on narrow viewports where the app bar's Library link is hidden.

- **Route handling**: The root route now redirects old `/` links with search or page params to `/library`, preserving bookmarked searches.

- **Tests**: Added comprehensive `AppBar.test.tsx` covering navigation, active states, drawer behavior on mobile, and account menu placement. Updated existing tests to use `/library` instead of `/`.

- **Moved/renamed**: `BookList` moved to `components/books/`, `PageHeader` docs updated to reflect its use on pages (not just tabs), `BookCard` moved alongside `BookList`.

The app bar is now the single place to reach any top-level destination from anywhere, avoiding conflicts with the book page's own tab and bottom navigation.

https://claude.ai/code/session_019Bvn1cY5PSES1zHX7pHcBg


Tumetsu:
The new global navigation has no active destination on book pages. `Library` uses fuzzy matching against `/library`, but book routes live under the sibling `/book/...` path, so both Home and Library render inactive while viewing a book. Since book pages explicitly return to—and conceptually belong to—the library, could the route configuration support an active-match alias/predicate for `/book/...`? It would also be useful to cover this state in the existing book-page AppBar test.

frontend/src/components/layout/AppBarNavLinks.tsx
```diff
@@ -0,0 +1,42 @@
...
+              color: 'primary.contrastText',
+              fontWeight: isActive ? 600 : 400,
+              // An underline rather than a filled pill: the bar is already a
+              // solid colour, and a second block of colour on it reads as a
+              // button to press rather than as where you are.
+              borderBottom: 2,
+              borderColor: isActive ? 'primary.contrastText' : 'transparent',
+              borderRadius: 0,
+              px: 1.5,
+              '&:hover': { borderColor: isActive ? 'primary.contrastText' : 'currentColor' },
```
Tumetsu: **Hover erases the active marker.**

`currentColor` resolves to the button's own `color`, which is `primary.contrastText` — the same value the active branch uses two lines up. So hovering an inactive link gives it exactly the active underline, and two links claim "where you are" at once. The comment on line 26 is the one this breaks.

A dimmer hover keeps the distinction:

```ts
'&:hover': {
  borderColor: isActive
    ? 'primary.contrastText'
    : (theme) => alpha(theme.palette.primary.contrastText, 0.4),
},
```

frontend/src/components/layout/AppBar.tsx
```diff
@@ -19,6 +21,7 @@ import { useAuth } from '../../context/AuthContext.tsx';
 export function AppBar() {
   const { logout } = useAuth();
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+  const [isDrawerOpen, setIsDrawerOpen] = useState(false);
```
Tumetsu, line 28: **The drawer survives a resize past `md`.**

`isDrawerOpen` is not tied to the breakpoint. Open the drawer on a phone, then rotate to landscape or widen the window past `md`, and the drawer stays over a page whose nav links are now in the bar. The backdrop click still gets you out, so it is recoverable rather than stuck.

```ts
const isWide = useMediaQuery(theme.breakpoints.up('md'));
useEffect(() => {
  if (isWide) setIsDrawerOpen(false);
}, [isWide]);
```

frontend/src/components/layout/AppBarNavLinks.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * The app's top-level destinations as a row of links in the app bar, for
+ * viewports with room for them. Below `md` the same list is in the drawer.
+ */
+export const AppBarNavLinks = () => {
+  const matchRoute = useMatchRoute();
+
+  return (
+    <Box sx={{ display: { xs: 'none', md: 'flex' }, alignItems: 'center', gap: 0.5, ml: 3 }}>
+      {APP_ROUTES.map((route) => {
+        const isActive = !!matchRoute({ to: route.to, fuzzy: route.fuzzy });
```
Tumetsu: **Two active-state mechanisms** (same in `AppBarDrawer.tsx:46`.)

`isActive` comes from `useMatchRoute`, but the `createLink`-wrapped components already emit `data-status="active"` — which is what `AppBar.test.tsx` asserts on. So the test covers TanStack's detection, not the branch that actually drives the styling.

Styling off `&[data-status='active']` would collapse the two and let `fuzzy` drop out of `appRoutes.ts` entirely. They agree today; this is about not keeping two sources of truth for one thing.

Optional — no bug either way.

frontend/src/components/layout/AppBarNavLinks.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * The app's top-level destinations as a row of links in the app bar, for
+ * viewports with room for them. Below `md` the same list is in the drawer.
+ */
+export const AppBarNavLinks = () => {
+  const matchRoute = useMatchRoute();
+
+  return (
+    <Box sx={{ display: { xs: 'none', md: 'flex' }, alignItems: 'center', gap: 0.5, ml: 3 }}>
```
Tumetsu: **No `nav` landmark.** This Box, and the drawer's `List` in `AppBarDrawer.tsx:44`, are plain containers. `component="nav"` on the wrapper gives screen readers a way to jump to global navigation — which matters more now that this is the app's only one.

Tumetsu (commented):
The split is sound and every check is green, but one navigation was left pointing at the old `/`: deleting a book now drops the reader on a dashboard instead of a book list.

## Checks

| Check | Result |
|---|---|
| `npm run type-check` | pass |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm test` | 179 passed (26 files) |
| `npm run duplication-check` | 1.0%, under the 1.05 threshold |

The one new clone jscpd reports is between the desktop and phone variants of the same assertion in `AppBar.test.tsx` — that pairing is the point, leave it.

## Fix before merge

**`frontend/src/pages/BookPage/BookTitle/BookEditDialog.tsx:37` — delete a book, land nowhere useful.** (Outside this diff, so no inline comment.)

```ts
navigate({ to: '/' });
```

Before this PR, `/` was the all-books list, so deleting a book put you back in the list minus that book. Now `/` is the dashboard: a recent-reads row and a "Browse the library" button. The reader deletes a book and gets no confirmation that the list changed. `to: '/library'` restores the old behaviour. Nothing covers this path — `useDeleteBook` has no test.

## What reads well

- `appRoutes.ts` as the single list both surfaces read — a destination cannot exist in the drawer and not the bar.
- The `beforeLoad` redirect at `/` keeps bookmarked `?search=&page=` links working, and `LibraryPage.test.tsx` pins it.
- `PullToRefresh.test.tsx` moved to `/library` with a docstring saying why the library, not the dashboard, is the page that makes a refetch visible — the rename did not silently change what the test proves.

The rest of the findings are inline.

Issue #863, opened 2026-09-15
M4.6 Label choice when creating a highlight in the reader


Tumetsu:
Part of #749 (M4 — Create highlights in the web reader). Size: M.

Split out of #752, which ships the selection popover with **Highlight** and **Cancel** only.

## What

The selection popover lists the book's labels (`GET /books/{book_id}/highlight-labels`).
Tapping a label chip creates the highlight in that label at once (one tap, as with
Kindle or Apple Books colour dots). The plain **Highlight** button still creates an
unlabeled one, and **Cancel** still closes the popover. The optimistic decoration is drawn
in the chosen label's colour.

Reuse the label chip from the highlights page: extract `LabelChip` out of
`frontend/src/pages/BookPage/navigation/HighlightLabelsList.tsx` into a shared
component used by both places.

## Gap: books with no labels

`POST /books/{book_id}/highlights` accepts only a `highlight_style_id` that belongs to
**this book** (`CreateHighlightFromLocatorUseCase._label`). The book list
(`LabelResolutionService.resolve_combination_labels`) lists only book-scoped
combination styles. So a book with an EPUB whose highlights never came from KOReader
has no styles, and the popover has no labels to offer.

Decide and implement one of these:

- Let the create endpoint accept a **global** label and find or create the book's
  combination style from its `device_color`/`device_style`, the way the KOReader
  upload does (`highlight_style_repository.find_or_create`). Then the popover falls
  back to `GET /highlight-labels/global` when the book has none.
- Or offer only the plain Highlight button in those books, and document the reason.

PR #864, opened 2026-09-15
M4.3: selection popover in the reader

### LINKED ISSUE #752: M4.3 Selection popover in the reader
Part of #749 (M4 — Create highlights in the web reader). Size: M.

Popover on selection: highlight with label choice, cancel. Optimistic
decoration; roll back on error. Reuse the label components from the highlights page.

Tumetsu:
Fixes #752 (M4.3, epic #749).

Selecting words in the web reader now opens a small toolbar below them with **Highlight** and **Cancel**. Highlight draws the passage at once and saves it through `POST /books/{id}/highlights` (M4.2). If the server refuses, the drawing is taken back and a message says why.

Label choice is split out to #863, by decision: it needs a backend change for books whose highlights never came from KOReader, which have no styles to pick from.

## Commits

1. **A popover over a selection, with Cancel.** The engine seam gains `clearSelection()`; `useEbookReader` exposes the selection. The popover is a non-modal `Popper` below the selection.
2. **Highlight, drawn at once, taken back on error.** `useHighlightCreation` holds a stand-in decoration per highlight in flight.

## Decisions worth a look

**A `Popper`, not a `Popover`.** A Popover's backdrop would lie over the book's iframe, so the selection handles could no longer be dragged, and its focus trap would pull focus out of the frame, which deactivates the selection. The paper also cancels `mousedown` for that reason. The cost: a keyboard-only reader cannot Tab into the toolbar.

**Any location report closes the popover, not only a page turn.** Measured on the real engine: a font-size change and a host resize report the same position several times while moving the selection's rectangle.

**The stand-in stays until the saved highlight can be drawn.** A drawn highlight needs both its locator and the book details. So after the POST the hook fetches `GET /highlights/{id}/locator`, merges it into the cached locator list (never refetching the whole list, which is `staleTime: Infinity` because placing every highlight is expensive), waits for book details to refetch, and only then removes the stand-in. A shell test pins that no set handed to the engine lacks both.

**Plain generated functions, not `useCreateHighlight().mutate(…, { onSuccess })`.** React Query v5 fires per-call callbacks only for the latest call, and two highlights in flight must each settle.

**Errors, by status:**

| Status | Message |
|---|---|
| 422 | "Couldn't find where this passage is in the book, so it wasn't highlighted. Try selecting a little more text." |
| 503 | "The book's file couldn't be read, so the highlight wasn't saved." |
| other | "Failed to save the highlight. Please try again." |

A saved highlight the server cannot place is not drawn and raises no error, which is the existing rule in `decorations.ts`.

## Try on a real device (not verified)

- **Touch:** a tap on Highlight may drop the book's selection before `click` fires. The create starts from the selection held in React state, but I could not test a phone.
- **Look:** I did not check the toolbar by eye.
- **A long mouse drag that autoscrolls a column** might report a location and close the popover mid-drag; the harness cannot drag.
- **Closing the reader mid-save:** a failure still shows its snackbar on the book page. The message is accurate; say if you'd rather drop it.

## Checks

- Frontend: lint, prettier, type-check, knip, jscpd (threshold ratcheted 1.05 → 0.74), full vitest suite (34 files, 406 tests).
- Mutation-checked: every rule above broken in the smallest way fails a test.

https://claude.ai/code/session_01GsausiLwAApbXCLB4mHtdC


Tumetsu:
Third commit, `a6ec8b8f`, fixes the two mobile bugs found in testing: a drag over words selected nothing and turned the page, and the popover disappeared as the finger came up.

Both are Readium's injected `ColumnSnapper`. It listens for `touchstart`/`touchmove`/`touchend` on the **frame's window** (passive, bubble phase). Its first `touchmove` calls `deselect()`, which removes the selection outright; its `touchend` snaps and reports progress, which arrives here as a location report — and a location report closes the popover, because a move leaves its rectangle behind.

The frame's document hears a touch before the window does, so `keepTouchFromTheSnapper` holds the gesture back in the capture phase **while the frame holds a selection**. The snapper never enters its drag, so it neither deselects nor reports. With nothing selected the gesture travels untouched, and a tap's collapsed caret counts as nothing selected, so letting the selection go hands page turns straight back. `preventDefault` is never called: the browser's own selection handles ride on the default.

**The cost, deliberate:** with words selected you cannot swipe to turn the page until you let the selection go by tapping.

Four tests dispatch real touch events at the live engine: a drag over a selection leaves it standing and reports no position; the same drag with nothing selected still reaches the page turner; so does one after `clearSelection()`; and so does one with only a caret. The first fails without the fix (`expected '' to be 'rarest and purest'`), and dropping the caret half of the guard fails three.

**Still only testable by hand:** iOS Safari (the harness is Chromium), and where the popover sits against the phone's own copy/paste menu.

Issue #866, opened 2026-09-16
A highlight that crosses a chapter boundary


Tumetsu:
Split out of the M4.7 discussion (#752's follow-up work). Size: L, and it changes a
contract KOReader also reads — do not start it without deciding the shape first.

A passage running from the end of one chapter into the next cannot be highlighted, and
the reason is structural rather than a missing button:

- **The browser:** each chapter is loaded into its own iframe, and a `Range` lives in one
  document. Two ranges would have to be stitched, and the browser's own selection UI
  cannot show one selection across both.
- **The locator:** `BrowserLocatorSchema` names one `href`, and
  `XPointCfiPositionAnchorService` resolves it inside a single resource
  (`_document(book, locator.href)`).
- **The canonical coordinates:** a highlight stores one `XPointRange`. Whether a start and
  an end in different spine documents is even expressible has to be settled against
  `xpoint-cfi` and against what KOReader draws when it syncs one back.
- **Everything downstream** reads the stored range: the chapter attribution, the position
  ordering, the drawn decoration, the e-reader.

## What a proposal has to answer

1. Can a KOReader xpointer range span two `DocFragment`s, and does KOReader render such a
   highlight? Check against the reader's own version, not master.
2. What the API takes: one locator with two resources, or two locators, or a highlight
   stored as two rows the views join.
3. What the web reader draws while the passage is half off-screen in another frame.
4. Whether the plugin's minimum version has to move with it (CLAUDE.md's rule for the
   plugin-only surface), and whether the sync can carry it at all.

Until then the reader refuses politely: a tap in another chapter does not extend a
selection (M4.7, #865).

PR #867, opened 2026-09-16
M4.7: extend a selection across pages

### LINKED ISSUE #865: M4.7 Extend a selection across pages in the reader
Part of #749 (M4 — Create highlights in the web reader). Size: M.

A reader can only highlight what fits on one page: a selection cannot be dragged past
the page edge, and the reader lets the selection go whenever the page moves
(`useEbookReader`'s location listener). A passage straddling a page boundary cannot be
highlighted at all.

Nothing below the UI is in the way. A reader page is a CSS column of one chapter
document, so a passage spanning pages is one ordinary `Range`, `selectionLocation`
already builds its locator, and the backend anchors a quote spanning several blocks.

## The flow, decided with the user

1. The reader selects the first words. The popover offers **Extend** beside Highlight
   and Cancel.
2. Extend remembers where the selection starts and lets the browser selection go, so
   swiping and the page-turn controls work as usual (the touch guard added in #752 only
   holds gestures back while a selection is up).
3. A bar says what to do — "Tap where the highlight ends" — with a way out.
4. The reader turns to the page they want and taps or clicks the last word. The engine
   builds the range from the remembered start to the tapped point
   (`caretPositionFromPoint`, with `caretRangeFromPoint` as the fallback), makes it the
   browser's selection so the words show as selected, and reports it as an ordinary
   selection. The popover comes back with Highlight and Cancel.

## Boundaries

- A tap in **another chapter** cannot extend the selection: each chapter is its own
  iframe and a `Range` lives in one document. Say so and keep the bar up. Crossing a
  chapter boundary is #866.
- Extending backwards (tapping before the start) should either work by swapping the
  ends or be refused; decide with a test either way.

## Seam

`EbookReader` gains `startSelectionExtension()`, `cancelSelectionExtension()` and a way
to report a tap that cannot extend. The extended passage arrives through the existing
`onSelectionChanged`, so the popover needs no new path for it.

Tumetsu:
Fixes #865 (M4.7, epic #749).

A passage that runs over a page break could not be highlighted: a drag cannot leave the visible column, and the reader lets the selection go whenever the page moves. Now the popover offers **Extend** — the reader turns pages as usual and taps where the passage ends.

Nothing below the UI stood in the way. A reader page is a CSS column of one chapter document, so a passage spanning pages is one ordinary `Range`, `selectionLocation` already builds its locator, and the backend anchors a quote spanning several blocks.

## Commits

1. **The engine remembers where a selection starts.** `startSelectionExtension()` keeps the start and lets the selection itself go, which hands gestures back: the touch guard only holds them while a selection is up, so swiping still turns pages. The next tap in that chapter becomes the far end.
2. **The reader's side of it.** Extend in the popover, a bar over the page while the tap is awaited, and the message when a tap cannot extend.

## Decisions worth a look

**A tap before the start flips the ends** rather than being refused. A reader who tapped slightly too early meant the passage, not an error.

**A tap in another chapter is refused, and the remembered start is kept,** so the reader can turn back and tap again. A `Range` lives in one document and each chapter is its own frame; #866 records what crossing that boundary would take on both sides.

**Every tap while an extension waits is the extension's, wherever it lands.** Readium activates a decoration from its own `pointerup` listener on the chapter, so a tap not stopped there opens a highlight underneath while the reader is still choosing where the passage ends. Found by review, with a test.

**The popover's rectangle is the part of the range on screen.** A bounding box over a passage crossing pages covers every page it crosses, which would put the popover off the page being read.

**The bar sits inside a region mounted for as long as the book is,** because a live region that appears together with its words announces nothing — the same lesson as the reconnecting notice above it.

**`listenerSet` collects the subscribe-and-notify shape** the engine repeated per event, semantics unchanged (the walk is over a copy, so a listener let go of mid-report still hears that one). The fake keeps its own sets: sharing the helper there makes the two classes textually converge, which measured worse (0.788% against 0.712%).

## Checks

Frontend lint, prettier, type-check, knip, jscpd (threshold ratcheted 0.74 → 0.72) and the full vitest suite. Each rule above is mutation-checked: broken in the smallest way, a test fails.

A page test carries the whole flow over the real engine — select, Extend, turn the page, tap, Highlight — and asserts the stored quote runs from the first words to the tapped ones.

## Please try on the phone

- Extend, swipe a page or two, then tap. Swiping should work while the bar is up, since the extension lets the browser selection go; that is reasoning, not something Chromium's harness proves.
- Whether the bar sits where you want it. It covers a strip of the first line at the top centre; the bottom belongs to your thumbs and the phone's own selection menu.
- A tap in the next chapter should say "A highlight has to stay inside one chapter." and leave the bar up.

https://claude.ai/code/session_01GsausiLwAApbXCLB4mHtdC


Tumetsu:
Two commits for the iOS Safari report: the passage flashed and vanished after tapping where it ends.

**Cause.** `extend` selects the words itself, and WebKit collapses a selection nobody dragged. That arrives as an ordinary `selectionchange` with nothing selected, and the settled report handed the UI a `null`, closing the popover over a passage the reader had just built. Desktop never collapses it, so the same report carried the same location and the dedupe swallowed it.

**`6f87186a` — the passage is the engine's own.** Which event asked for a report now decides what it means: `selectionchange` settling with nothing selected is the browser talking about itself, while a pointerup is the reader saying they are done. A held passage is released by a tap, a new selection, `clearSelection()` or `destroy()` — never by the browser alone. `EbookSelection` carries whether the browser still shows the words, and the dedupe key carries it too, so "report it again, once" falls out of the existing dedupe. The reader draws a passage the browser has dropped, since on iOS nothing else marks the words the popover is about.

**`a790fe16` — a second Extend on a held passage.** Extend reads its anchor from a live browser selection, of which there is none once WebKit has collapsed one, so pressing Extend again would have waited for a tap that could never complete. The start a further extension runs from now lives inside the held passage, so every release the passage already has drops the start with it.

**One harness change worth a look.** The tests dispatched a tap on the frame's document, where the extension's capture-phase `stopPropagation` cannot stop a bubble listener on that same node — so the tap also reported through the ordinary path and would have released the hold on the spot. A real tap targets an element, where propagation does stop. The harness now does the same, rather than production code carrying a rule that exists only for a synthetic event path.

**Unverified here:** that WebKit's collapse is what fires `selectionchange`, and whether it lands inside the 200 ms settle window. The fix is indifferent to both — the hold survives any number of empty `selectionchange`s — but only a device settles it.

Tumetsu:
`45a62f23` carries three fixes from device testing, **implemented without new tests** at the user's request. The existing suite still passes unchanged (212 reader tests, lint, prettier, type-check, knip, jscpd); CI runs the rest.

**The reader stuck mid-turn.** Reproduction: select words, press Extend, then swipe. The touch guard decided per event, so a swipe beginning with nothing selected was taken by Readium's `ColumnSnapper` and then cut off the moment WebKit created a selection part-way through the drag. The snapper runs a state machine over `touchstart`/`touchmove`/`touchend`; handed only a gesture's head it was stranded mid-transition, and every later gesture was stopped too, because a selection existed by then. The decision is now taken once per gesture and held to its end, with `touchcancel` added, so the snapper sees a whole gesture or none of it.

**The bar gone, the popover up.** That stray mid-drag selection was reported as an ordinary one, so the bar lowered while the engine still held the anchor — the next tap would have run a passage from a forgotten place. Words selected while an extension waits now take the anchor with them, whichever event reported them.

**A passage ending mid-word.** A tap resolves to a character position. Both ends of the range now grow out to whole words — both, because the flipped case means either end may be the one the tap set — and only when the boundary sits strictly inside a word, so an edge that was already right stays put. Words are letters and digits, which leaves punctuation stuck to a word outside it.

**Coverage gap worth knowing:** the per-gesture rule has no test of its own. It is the failure that leaves the reader unusable, and the harness cannot reproduce WebKit creating a selection mid-drag, but a synthetic version (gesture starts clean, a selection is made between two moves, the snapper must still act) would pin the regression. Say the word and I will add it.

